PR #5, opened 2019-06-24
multi-root workspace support

cmatthiassonsf:


packages/lightning-lsp-common/src/context.ts
```diff
@@ -320,18 +323,18 @@ export class WorkspaceContext {
     }
 
     private async readConfigBlt() {
```
nrkruk: The logic in this method is confusing - probably could be cleaned up

packages/lightning-lsp-common/src/context.ts
```diff
@@ -357,25 +360,27 @@ export class WorkspaceContext {
      */
     private updateConfigFile(relativeConfigPath: string, config: string) {
```
nrkruk: not sure if this is exactly right either - need to test it locally

packages/lwc-language-server/src/config.ts
```diff
@@ -15,11 +16,11 @@ export interface IJsconfig {
         paths?: IPaths;
     };
 }
-
+// this listens and i have to detect where this file is coming from
 export async function onIndexCustomComponents(context: WorkspaceContext, files: string[]) {
     // set paths for all current components in all the projects jsconfig.json files
     for (const relativeModulesDir of await context.getRelativeModulesDirs()) {
-        const modulesDir = path.join(context.workspaceRoot, relativeModulesDir);
+        const modulesDir = path.join(context.workspaceRoots[0], relativeModulesDir);
```
nrkruk: Logic in this method looks wrong to me now - I think we need to get rid of this getRelativeModulesDir() method and just have that method return the full list of absolute paths to module directories, rather than getting relative ones and having to do path joins. That will simplify this greatly.

packages/lwc-language-server/src/metadata-utils/content-assets-util.ts
```diff
@@ -20,7 +20,7 @@ function getResourceName(resourceMetaFile: string) {
     return parse(resourceFile).name;
 }
 
-export async function updateContentAssetIndex(updatedFiles: FileEvent[], { workspaceRoot }: WorkspaceContext, writeConfigs: boolean = true) {
+export async function updateContentAssetIndex(updatedFiles: FileEvent[], { workspaceRoots }: WorkspaceContext, writeConfigs: boolean = true) {
```
nrkruk: All this wouldn't work in a multi root scenario but thats probably OK. I think we just need to document which methods are used explicitly for SFDX projects vs which ones are used for core / internal development

packages/lightning-lsp-common/src/context.ts
```diff
@@ -200,13 +211,13 @@ export class WorkspaceContext {
 
         switch (this.type) {
             case WorkspaceType.SFDX:
-                typingsDir = join(this.workspaceRoot, '.sfdx', 'typings', 'lwc');
+                typingsDir = join(this.workspaceRoots[0], '.sfdx', 'typings', 'lwc');
                 break;
-            case WorkspaceType.CORE_SINGLE_PROJECT:
-                typingsDir = join(this.workspaceRoot, '..', '.vscode', 'typings', 'lwc');
+            case WorkspaceType.CORE_PARTIAL:
+                typingsDir = join(this.workspaceRoots[0], '..', '.vscode', 'typings', 'lwc');
```
nrkruk: How does this work in core? (Nick TODO)

Issue #15, opened 2019-08-27
Add support for Lightning Web Components Open Source


muenzpraeger:
**Is your feature request related to a problem? Please describe.**
Currently LWC OSS is not supported in the language server. This reduces the developer experience when working with that technology outside of Salesforce Platform.

**Describe the solution you'd like**
Support for LWC OSS should be implemented.

**Describe alternatives you've considered**
None

PR #16, opened 2019-08-27
Add support for LWC off-platform

muenzpraeger:
### What does this PR do?
This PR adds support for LWC OSS.

#15


packages/lwc-language-server/src/javascript/compiler.ts
```diff
@@ -91,7 +93,33 @@ export async function compileSource(source: string, fileName: string = 'foo.js')
             namespace: 'x',
             files: {},
         };
-        const transformerResult = await transform(source, fileName, options);
+
+        let codeTransformed = source;
+
+        if (fileName.endsWith('.ts')) {
+            try {
+                const { code } = babel.transform(source, {
```
KrisGraySFDC: Lets extract this into a different method.

packages/aura-language-server/src/server.ts
```diff
@@ -234,6 +250,10 @@ connection.onDefinition(
 );
 
 connection.onDidChangeWatchedFiles(async (change: DidChangeWatchedFilesParams) => {
+    if (isLwcOss) {
```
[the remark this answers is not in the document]
muenzpraeger: Because the Aura server always starts and does his stuff also in an LWC OSS project. Which then throws errors I. The Aura LS.

packages/lightning-lsp-common/src/context.ts
```diff
@@ -271,6 +290,23 @@ export class WorkspaceContext {
                     this.updateConfigFile(relativeJsConfigPath, jsConfigContent);
                 }
                 break;
+
+            case WorkspaceType.OSS_LWC:
+                jsConfigTemplate = await fs.readFile(utils.getCoreResource('jsconfig-core.json'), 'utf8');
+                const fileName = path.resolve(process.cwd(), 'lwc-services.config.js');
```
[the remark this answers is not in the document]
muenzpraeger: Yup, I basically copied the way it’s done with the other workspace types.

packages/lwc-language-server/src/javascript/compiler.ts
muenzpraeger: Extracted.

packages/lightning-lsp-common/src/__tests__/context.test.ts
```diff
@@ -59,6 +68,16 @@ it('WorkspaceContext', async () => {
     expect(modules[0]).toEndWith(join(CORE_PROJECT_ROOT, 'modules', 'one', 'app-nav-bar', 'app-nav-bar.js'));
     expect(modules.length).toBe(1);
     expect(await context.getRelativeModulesDirs()).toEqual(['modules']);
+
+    context = new WorkspaceContext(OSS_LWC_PROJECT_ROOT);
```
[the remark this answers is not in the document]
muenzpraeger: Done.

packages/aura-language-server/src/server.ts
```diff
@@ -78,6 +80,14 @@ connection.onInitialize(
                 return { capabilities: {} };
             }
 
+            const lwcServicesConfig = path.resolve(workspaceRoot, 'lwc-services.config.js');
+
+            if (lwcServicesConfig) {
+                console.info(`LWC Open Source project, Aura language server not needed.`);
+                isLwcOss = true;
```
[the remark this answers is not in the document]
muenzpraeger: Actually I think I found a better way - not starting the Aura LS at all. This is the "[problematic](https://github.com/forcedotcom/salesforcedx-vscode/blob/develop/packages/salesforcedx-vscode-lightning/src/index.ts#L73)" check in the VS Code extension itself.

I added an `isAura` function [here](https://github.com/forcedotcom/lightning-language-server/pull/16/files#diff-b7cf3b6b041b8f0e826428e4615fbe5fR48) to the language server, which excludes LWC OSS. And updated the check as in [here](https://github.com/forcedotcom/salesforcedx-vscode/pull/1655/files#diff-57e4c3399a2262cbfe3bcb708b22acf8R75).

Would need some cross-coordination with @lcampos to get the changes into the same release. ;-)

muenzpraeger:
Just bumping this up... @KrisGraySFDC

packages/lwc-language-server/src/javascript/compiler.ts
```diff
@@ -91,7 +93,16 @@ export async function compileSource(source: string, fileName: string = 'foo.js')
             namespace: 'x',
             files: {},
         };
-        const transformerResult = await transform(source, fileName, options);
+
+        let codeTransformed = source;
+
+        if (fileName.endsWith('.ts')) {
+            codeTransformed = transformTypescript(fileName, source);
```
lcampos, line 100: I don't think this is a good idea. In case of errors we wouldn't be able to provide accurate information that maps to the line & column of the component file(s).

muenzpraeger: Good catch - I think applying this should fix it.


https://github.com/salesforce/lwc-test/pull/75

xyc: Is there a reason why lwc compiler cannot support `.ts`?

muenzpraeger: Well, good question to ask the LWC team. As of now it's not.

PR #19, opened 2019-11-27
[W-6860946] Reduce writes to jsconfig.json

bpbuch:
### What does this PR do?
* Added onDidSave event handler to LWC server connection.
* Modified writeConfigs to reduce config update attempts.
* Added a check for differences prior to writing to config file.

W-6860946

### DIFF
--- a/packages/lwc-language-server/src/config.ts
+++ b/packages/lwc-language-server/src/config.ts
@@ -1,6 +1,7 @@
 import * as path from 'path';
 import { WorkspaceContext, shared, utils, componentUtil } from 'lightning-lsp-common';
 import { readJsonSync, writeJsonSync } from 'lightning-lsp-common/lib/utils';
+import equal from 'deep-equal';
 
 const { WorkspaceType } = shared;
 
@@ -55,7 +56,7 @@ export async function onIndexCustomComponents(context: WorkspaceContext, files: 
     }
 }
 
-export async function onCreatedCustomComponent(context: WorkspaceContext, file: string): Promise<void> {
+export async function onSetCustomComponent(context: WorkspaceContext, file: string): Promise<void> {
     if (!file) {
         // could be a non-local tag, like LGC, etc
         return;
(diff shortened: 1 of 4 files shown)


packages/lwc-language-server/src/config.ts
```diff
@@ -74,15 +75,22 @@ export async function onCreatedCustomComponent(context: WorkspaceContext, file:
...
+
             if (!jsconfig.compilerOptions.paths) {
-                jsconfig.compilerOptions.paths = {};
+                newJsconfig.compilerOptions.paths = {};
+            }
+
+            newJsconfig.compilerOptions.baseUrl = '.';
+            newJsconfig.compilerOptions.paths[moduleTag] = [relativeFilePath];
+
+            if (!equal(jsconfig, newJsconfig)) {
```
KrisGraySFDC: Do we have perf metrics to tell us if this added any cost? I imagine the files are small enough and this action doesn't happen that often, but curious still

packages/lwc-language-server/src/metadata-utils/custom-components-util.ts
```diff
@@ -120,7 +120,9 @@ async function addCustomTag(context: WorkspaceContext, tag: string, uri: string,
 export async function indexCustomComponents(context: WorkspaceContext, writeConfigs: boolean = true): Promise<void> {
     const files = await context.findAllModules();
 
-    await loadCustomTagsFromFiles(context, files, context.type === WorkspaceType.SFDX, writeConfigs);
+    // writeConfigs is set to false to avoid updating config twice for the same tag.
```
KrisGraySFDC: Nice comment.

KrisGraySFDC: Would be nice to refactor to avoid the boolean trap.

packages/lwc-language-server/src/config.ts
bpbuch: The instrumentation for metrics along this code path is not yet implemented, but I agree that this action should be relatively trivial.  It could be an interesting experiment to see the performance between allowing and blocking an identical write.

PR #25, opened 2020-01-21
add tests for package.json

mysticflute:
add test to check that dependencies on @salesforce scoped or other packages in the monorepo use exact version ranges (no ^ or ~).

also updates jest dependencies, adds jest-extended, updates a few package version ranges and other minor cleanup.

@W-7129288@


packages/aura-language-server/jest.config.js
```diff
@@ -9,6 +9,6 @@ module.exports = {
       "js",
       "json"
     ],
-    setupTestFrameworkScriptFile: "<rootDir>/jest/matchers.ts",
+    setupFilesAfterEnv: ["jest-extended"],
```
KrisGraySFDC: Just asking about this as I'm not sure whats going on.

mysticflute: I upgraded the version of jest and this is the new way to specify `setupTestFrameworkScriptFile`, as this one takes more than one file. jest-extended is an extended set of matchers, one of which I used in my test and another one which replaces a custom matcher we were using

packages/aura-language-server/src/__tests__/package-dependencies.test.ts
```diff
@@ -0,0 +1,83 @@
+import path from 'path';
+import fs from 'fs';
+import glob from 'glob';
+
+// These unit tests check that specified dependencies in package.json do not use
+// ^ or ~ in the version range, either because those packages do not use semver
+// and minor/patch updates will break functionality, or because they should only
+// use an exact version.
+
+const checkedPackagePatterns: RegExp[] = [/^@salesforce/i, /^@lwc/i];
```
mysticflute: I added @lwc here as well

PR #26, opened 2020-01-24
[W-7151419] Don't update jsconfig with existing content

bpbuch:
### What does this PR do?

Added a check to verify if the jsconfig content needs to be updated.

@W-7151419@

### DIFF
--- a/packages/lwc-language-server/src/config.ts
+++ b/packages/lwc-language-server/src/config.ts
@@ -33,6 +33,8 @@ export async function onIndexCustomComponents(context: WorkspaceContext, files: 
         try {
             // note, this read/write file must be synchronous, so it is atomic
             const jsconfig: IJsconfig = readJsonSync(jsconfigFile);
+            // deep clone of jsconfig created for update comparison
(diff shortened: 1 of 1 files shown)


packages/lwc-language-server/src/config.ts
```diff
@@ -33,6 +33,8 @@ export async function onIndexCustomComponents(context: WorkspaceContext, files:
         try {
             // note, this read/write file must be synchronous, so it is atomic
             const jsconfig: IJsconfig = readJsonSync(jsconfigFile);
+            // deep clone of jsconfig created for update comparison
+            const newJsconfig: IJsconfig = JSON.parse(JSON.stringify(jsconfig));
```
nrkruk: Worried about the performance implications of this on core. We'll need to do some testing to see

bpbuch: Agreed.  We need to refactor some of the indexing to avoid making it this far.  There are several actions that lead to a config update that probably shouldn't.

For now, I don't think the performance will be too terribly impacted.  The config updates aren't very frequent now that it doesn't happen on every keystroke.  The biggest factor would be how large this file becomes for mature projects.

PR #39, opened 2020-02-06
Add messageChannel support to the lwc language server

nrkruk:
### What does this PR do?
Content assist on @salesforce/messageChannel


packages/lwc-language-server/src/metadata-utils/message-channel-util.ts
```diff
@@ -0,0 +1,81 @@
...
+        throw err;
+    }
+}
+
+
+
+function generateTypeDeclarations(): string {
+    let resTypeDecs = '';
+    const sortedContentAssets = Array.from(MESSAGE_CHANNELS).sort();
+    sortedContentAssets.forEach(res => {
```
KrisGraySFDC: Isn't this a map and a join?

return sortedContentAssets.map(generateTypeDeclaration).join('');

package.json
```diff
@@ -2,6 +2,7 @@
   "scripts": {
     "bootstrap": "lerna bootstrap --force-local",
     "clean": "lerna run clean --stream",
+    "cleanTestData": "lerna run cleanTestData --stream",
```
bpbuch: Small Nit: the scripts are inconsistent in their naming convention.  There are scripts with: camel case, snake case, and gulp style.

trentjohn:
Generally, the impl looks good. It uses the proper syntax for @salesforce/messageChannel and has the right inline documentation for the latest and greatest exports from `lightning/messageService`

@nrkruk confirms this iteration doesn't parse the description out of the XML, to be shown in the Intellisense popup. Requesting that functionality per the [W-6744593](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07B0000007Zn9xIAC/view) description.

image
image

PR #40, opened 2020-02-07
[W-7158536] Add dependabot config

bpbuch:
### What does this PR do?
Add Dependabot config
@W-7158536@

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -1,0 +1,10 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "daily"
+    allowed_updates:
+      - match:
+          dependency_name: "@lwc/*"
+      - match:
+          dependency_name: "@salesforce/*"


.dependabot/config.yml
```diff
@@ -0,0 +1,10 @@
+version: 1
```
nrkruk: does this need to be in our packages specifically?

bpbuch:
```
The config file must start with version: 1 followed by an array of update_configs.
```

bpbuch: No. It can be at the root and watch all nested packages.

nrkruk: oh this is the language server, duh

.dependabot/config.yml
```diff
@@ -0,0 +1,10 @@
...
+update_configs:
+  - package_manager: "javascript"
+    directory: "/"
+    update_schedule: "daily"
+    allowed_updates:
```
ntotten: Does this mean it will only send updates for our own stuff. Wouldn't it be better to do updates for anything, i.e. the vscode language server also?

mysticflute:
@bpbuch is there some way to send group PRs for everything under `@lwc`?  I think the individual PRs will cause CI problems since there could be incompatible changes sometimes

bpbuch:
Not that I am aware of. There is an open issue for others requesting this as a feature.

https://github.com/dependabot/dependabot-core/issues/1296

.dependabot/config.yml
bpbuch: I am open to having dependabot watch all dependencies.  I believe @nrkruk had some reservations about this.

PR #45, opened 2020-02-10
Publish to npmjs

KrisGraySFDC:
### What does this PR do?
Start publishing to npmjs vs lwcjs

@W-6819905


.circleci/config.yml
```diff
@@ -5,10 +5,13 @@ orbs:
   codecov: codecov/codecov@1.0.5
   browser-tools: circleci/browser-tools@0.1.3
 
+defaults: &defaults
```
mysticflute: just curious what this means?

.circleci/config.yml
```diff
@@ -29,3 +32,33 @@ jobs:
...
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+      - run: yarn install
+      - run: yarn build
+      - run:
+          name: Authenticate with NPM
+          command: 'echo "//registry.npmjs.org/:_authToken=$NPM_PUBLISH_TOKEN" > ~/lightning-language-server/.npmrc'
+      - run:
+          name: Publish to NPM
+          command: yarn publish-lsp
```
nrkruk: I think we need to commit changes after this right because the versions will get bumped during this process

.circleci/config.yml
```diff
@@ -29,3 +32,33 @@ jobs:
           key: dependency-cache-{{ checksum "package.json" }}
           paths:
             - node_modules
+
+  # TODO: Don't recompile everything. Use workspace persistance.
+  deploy:
```
nrkruk: Does this run on every branch or just master?

nrkruk: nevermind I see below

.circleci/config.yml
KrisGraySFDC: Its a variable block, something you can reference later using the <<: *[name] syntax

.circleci/config.yml
KrisGraySFDC: Lerna does it automatically.

PR #47, opened 2020-02-11
fix issues with unit tests giving false positives

nrkruk:
### What does this PR do?
Unit tests were blindly passing due to async refactor. Fixes unit tests to await async indexing then check the actual file system for typescript definition files.


packages/lightning-lsp-common/src/resources/sfdx/typings/messageservice.d.ts
```diff
@@ -16,9 +16,9 @@ declare module 'lightning/messageService' {
      * @param {Function} listener - Function to be invoked when messages are published on the channel.
      * @param {Object} subscriberOptions - Options to influence message channel subscription.
      *                                     Current subscriber options:
-     *                                       1. `scope` - the scope that a component is subscribed to.
```
mysticflute: ?

packages/lwc-language-server/src/metadata-utils/__tests__/content-assets-util.test.ts
```diff
@@ -1,12 +1,17 @@
...
-// @ts-ignore
-import { WorkspaceContext } from 'lightning-lsp-common';
+import { FORCE_APP_ROOT } from '../../__tests__/test-utils';
+import * as fs from 'fs-extra';
 
-jest.mock('lightning-lsp-common', () => {
-    const real = jest.requireActual('lightning-lsp-common');
-    return { utils: mockFileUtil(), WorkspaceContext: real.WorkspaceContext };
+beforeEach(() => {
+    const jsconfigPathForceApp = FORCE_APP_ROOT + '/lwc/jsconfig.json';
```
mysticflute: if this has to happen before each unit test can we extract it to a shared function or something

PR #48, opened 2020-02-11
Windows circleci2

KrisGraySFDC:
### What does this PR do?
Runs the build on CircleCI for windows vs Appveyor.
Appveyors config file has been deleted in this change.

@W-7124411


packages/lwc-language-server/package.json
```diff
@@ -52,8 +52,10 @@
     "@types/glob": "^5.0.35",
     "@types/jest": "^25.1.2",
     "@types/node": "^8.10.0",
+    "@types/utf8": "^2.1.6",
```
KrisGraySFDC: Get rid of this.

.circleci/config.yml
```diff
@@ -2,23 +2,64 @@
...
 
 orbs:
+  # Code Coverage
   codecov: codecov/codecov@1.0.5
+  # For Installing Browsers
   browser-tools: circleci/browser-tools@0.1.3
+  # For Testing on windows
+  win: circleci/windows@2.4.0
 
 defaults: &defaults
```
KrisGraySFDC: Think we can get rid of this now too.

.circleci/config.yml
```diff
@@ -52,12 +101,19 @@ jobs:
...
           requires:
             - build
+      - tests-on-windows:
+          requires:
+            - build
+      - deploy:
+          requires:
+            - tests-with-coverage
+            - tests-on-windows
           filters:
```
nrkruk: We don't need the branch filters for build and test right? We just want the filters to apply to deploying

KrisGraySFDC: Yea thats what I thought this said.
Deploy requires these two builds and only runs on master.

.circleci/config.yml
```diff
@@ -2,23 +2,64 @@
...
+      # Cache the node_modules folder so if package.json doesn't change
+      # then we can prevent downloading it all over again.
+      - save_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - node_modules
+
+      # Save the current workspace state so the next jobs can
+      # re-use it without having to do the build over again.
+      - persist_to_workspace:
```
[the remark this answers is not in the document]
KrisGraySFDC: Deploy step still wants this though. Otherwise we'll need to rebuild for that to deploy the output.

.circleci/config.yml
```diff
@@ -2,23 +2,64 @@
 version: 2.1
```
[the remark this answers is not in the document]
KrisGraySFDC: And deploy

.circleci/config.yml
```diff
@@ -28,35 +59,50 @@ jobs:
...
           key: dependency-cache-{{ checksum "package.json" }}
+      - attach_workspace:
+          at: ~/lightning-language-server
       - run: yarn install
       - run: yarn build
-      - run: git config user.email "nkruk@salesforce.com"
-      - run: git config user.name "Nick Kruk"
+      - run: git config user.email "kgray@salesforce.com"
+      - run: git config user.name "Kris Gray"
+      - run: npm config set '//registry.npmjs.org/:_authToken' "${NPM_PUBLISH_TOKEN}"
```
KrisGraySFDC: Doing this to avoid having it in the NPMRC file. 

That is problematic for 2 reasons.

If you just put the Environment variable name in there, building fails unless you declare that value on the build host.
If I don't check that in, and do it at deploy time, lerna complains we have a changed file when its trying to do some stuff around pushing versions back to git.

Will verify this once it gets back to master.

.circleci/config.yml
```diff
@@ -2,23 +2,54 @@
...
+      - save_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
+          paths:
+            - node_modules
+
+      # Save the current workspace state so the next jobs can
+      # re-use it without having to do the build over again.
+      - persist_to_workspace:
+          root: ~/lightning-language-server
+          paths:
```
mysticflute: might be a maintenance annoyance going forward. how much longer is it really to have the other job rebuild only?

packages/lwc-language-server/src/metadata-utils/__tests__/util.ts
```diff
@@ -24,6 +25,8 @@ export async function validate(
     await indexer(context, true);
     const path = join(workspacePath, '.sfdx', 'typings', 'lwc', expectedTypeDeclarationFileName);
     expect(path).toExist();
-    const contents = fs.readFileSync(path, 'utf8');
-    expect(contents).toBe(expectedTypeDeclarations);
+    // For windows we need to normalize line endings, we do that using eol.
+    const contents = eol.auto(fs.readFileSync(path, 'utf8'));
```
mysticflute: seems weird to me that we have to use this eol thing. is it a problem we can fix with the file itself?

.circleci/config.yml
KrisGraySFDC: You mean they should all just checkout, install, build before doing their work?

packages/lwc-language-server/src/metadata-utils/__tests__/util.ts
KrisGraySFDC: As soon as someone goes and edits the file again, it'll all blow up. Line endings are a pain in the butt.

.circleci/config.yml
mysticflute: that just seems simpler than having to remember to go here and add new paths if we add a new folder or file later. if it's notably faster though then I'm ok with it

KrisGraySFDC: It'll be faster, we also shouldn't need any new stuff in here. Most things should be in the packages directory.

packages/lwc-language-server/src/metadata-utils/__tests__/util.ts
nrkruk: I think this is due to the workspace getting restored from cache, which would cause git to have checked out the filesystem with the wrong style line endings

nrkruk: we'll remove this later, just wanna get this PR in

PR #112, opened 2020-04-27
feat: 🎉 lightning namespace check

muenzpraeger:
### What does this PR do?
It adds additional validation if someone mis-types "lightning".

### What issues does this PR fix or reference?
#107

PR #113, opened 2020-04-28
feat: 🚀 add QuickFix actions

muenzpraeger:
### What does this PR do?
This PR will add QuickFix support for common LWC compiler errors.


muenzpraeger:
Includes/needs #112

PR #120, opened 2020-05-06
activate on the lwc-services dependency

miwialex:
### What does this PR do?

We're now checking the following cases to determine if a workspace
should be STANDARD_LWC:

- A `lwc.config.js` file exists at the root
- The `package.json` at the root has one of the following keys in
  `dependencies` or `devDepencencies`
  - starts with `@lwc/`
  - `lwc-services`
- The `package.json` has a root key (configuration) of `lwc`

### DIFF
--- a/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
@@ -87,6 +87,42 @@ describe('detectWorkspaceType', () => {
         expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
     });
 
+    test('when package.json dependencies include lwc-services, workspaceType is STANDARD_LWC', () => {
+        mockFs({
+            workspacedir: {
+                'package.json': JSON.stringify({
+                    dependencies: {
+                        'lwc-services': 1,
+                    },
+                }),
+            },
+        });
+
+        const workspaceType = detectWorkspaceType(['workspacedir']);
+
+        mockFs.restore();
+
+        expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
+    });
+
+    test('when package.json devDependencies include lwc-services, workspaceType is STANDARD_LWC', () => {
+        mockFs({
+            workspacedir: {
+                'package.json': JSON.stringify({
+                    devDependencies: {
+                        'lwc-services': 1,
+                    },
+                }),
+            },
+        });
+
+        const workspaceType = detectWorkspaceType(['workspacedir']);
+
+        mockFs.restore();
+
+        expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
+    });
+
     test('when package.json specifies workspaces, workspaceType is MONOREPO', () => {
         mockFs({
             workspacedir: {
--- a/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
@@ -30,6 +30,18 @@ describe('detectWorkspaceType', () => {
         expect(workspaceType).toEqual(WorkspaceType.SFDX);
     });
 
+    test('when an lwc.config.js file is present, workspaceType is STANDARD_LWC', () => {
+        mockFs({
+            workspacedir: {
+                'lwc.config.js': '',
+            },
+        });
+
+        const workspaceType = detectWorkspaceType(['workspacedir']);
+
+        expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
+    });
+
     test('when workspace-user.xml file is present at the root, workspaceType is CORE_ALL', () => {
         mockFs({
             workspacedir: {
--- a/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
@@ -112,8 +112,6 @@ describe('detectWorkspaceType', () => {
 
         const workspaceType = detectWorkspaceType(['workspacedir']);
 
-        mockFs.restore();
-
         expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
     });
 
@@ -129,8 +127,6 @@ describe('detectWorkspaceType', () => {
         });
 
         const workspaceType = detectWorkspaceType(['workspacedir']);
-
-        mockFs.restore();
 
         expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
     });
--- a/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
@@ -143,7 +143,7 @@ describe('detectWorkspaceType', () => {
         expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
     });
 
-    test('when package.json has any dependency that starts with "@lwc"', () => {
+    test('when package.json has the `lwc-services` dependency', () => {
         mockFs({
             workspacedir: {
                 'package.json': JSON.stringify({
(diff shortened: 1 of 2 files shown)


packages/lightning-lsp-common/src/shared.ts
```diff
@@ -73,25 +73,44 @@ export function detectWorkspaceHelper(root: string): WorkspaceType {
...
     const packageJson = path.join(root, 'package.json');
     if (fs.existsSync(packageJson)) {
         try {
-            // Check if package.json contains @lwc/engine
             const packageInfo = JSON.parse(fs.readFileSync(packageJson, 'utf-8'));
             const dependencies = Object.keys(packageInfo.dependencies || {});
-            if (dependencies.includes('@lwc/engine')) {
+            const devDependencies = Object.keys(packageInfo.devDependencies || {});
+            const allDependencies = [...dependencies, ...devDependencies];
+            const hasLWCdependencies = allDependencies.some(key => key.includes('@lwc/'));
```
xyc, line 87:
```suggestion
            const hasLWCdependencies = allDependencies.some(key => key.startsWith('@lwc/'));
```
Maybe we could use `startsWith`? Since we know that the pattern is in the beginning

packages/lightning-lsp-common/src/shared.ts
```diff
@@ -73,25 +73,44 @@ export function detectWorkspaceHelper(root: string): WorkspaceType {
...
-            if (devDependencies.includes('@lwc/engine')) {
+
+            // has and type of lwc configuration
+            if (packageInfo.lwc) {
                 return WorkspaceType.STANDARD_LWC;
             }
+
+            // has a lwc-services dependency (this ca be removed once the
+            // `lwc.config.js` becomes required
+            if (allDependencies.includes('lwc-services')) {
```
xyc, line 101: Are we considering to merge "Add Support for LWC OSS" PR #16 by @muenzpraeger ? Looks like that PR does some more language support scenarios specific to apps created by `create-lwc-app`. It also checks `lwc-services` [here](https://github.com/forcedotcom/lightning-language-server/pull/16/files#diff-b7cf3b6b041b8f0e826428e4615fbe5fR92-R94).

@muenzpraeger I also wonder if we are going to make a distinction in the future between:
- Generic LWC OSS projects

Quoting @mysticflute: (The latter ones)
> Some OSS users will use create-lwc-app , and apps created for demos, but internal off-core and core projects won't be using it. They will either have a dependency on @lwc, or have an lwc.config.json or have lwc config in the package.json
> Here's an example of an off-core project that is using the lwc config https://github.com/salesforce/lightning-components/blob/master/ui-lightning-components/package.json#L111
- Apps created by `create-lwc-app`

For now, it would help with extension capabilities (like determining a proper test runner for `create-lwc-app` apps)

xyc: I think @lcampos had mentioned the PR #16 needs some rework, maybe we can leave this be now.

packages/lightning-lsp-common/src/shared.ts
```diff
@@ -73,25 +73,44 @@ export function detectWorkspaceHelper(root: string): WorkspaceType {
...
+            const hasLWCdependencies = allDependencies.some(key => key.includes('@lwc/'));
+
+            // any type of @lwc is a dependency
+            if (hasLWCdependencies) {
                 return WorkspaceType.STANDARD_LWC;
             }
-            const devDependencies = Object.keys(packageInfo.devDependencies || {});
-            if (devDependencies.includes('@lwc/engine')) {
+
+            // has and type of lwc configuration
```
xyc, line 94: typo and -> any?

packages/lightning-lsp-common/src/shared.ts
muenzpraeger: It should also just check for the `lwc` package. 

https://www.npmjs.com/package/lwc

packages/lightning-lsp-common/src/shared.ts
muenzpraeger: There are a couple of things, also based on a convo with @mysticflute:

- I will add support to create-lwc-app to use `lwc.config.js` - so we won’t need the additional check
- The module resolution mechanism, and the shape for the config of `lwc.config.js` has a breaking change after/with 1.4.0.

Now, coming back to no 1, I think we can remove the lwc-services check.

For no 2 there needs to be a decision. Latest with 228 LWC 1.4.0 will be internally also usable/used. In OSS (and likely many off core projects it’s also used). I am happy to put in a new PR, based on this PR, who makes usage of the new module resolution. 

One caveat, in general, is that the LWC versions for on and off platform  deviate. I would propose to update to 1.5.x. for the tooling here in general.

Thoughts @lcampos @diervo ?

packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
```diff
@@ -81,8 +140,68 @@ describe('detectWorkspaceType', () => {
         });
 
         const workspaceType = detectWorkspaceType(['workspacedir']);
+        expect(workspaceType).toEqual(WorkspaceType.STANDARD_LWC);
+    });
 
-        mockFs.restore();
+    test('when package.json has any dependency that starts with "@lwc"', () => {
```
muenzpraeger, line 146: Test description doesn't map to test.

muenzpraeger:
I’m pushing an update to #16, wich should supersede this one.

muenzpraeger:
@xyc I pushed an update to #16, based on how module resolution works for LWC 1.4.0 onwards, and which includes the elements of this PR. Note that it'll only provide language features for local LWC folders., which should be sufficient atm.

This may need to be extended in the future, in case there are future plans for adding language support for packages from npm for example. There's a whole package that @diervo built [here](https://github.com/salesforce/lwc/tree/master/packages/%40lwc/module-resolver), which helps to resolve, but not to traverse.

packages/lightning-lsp-common/src/shared.ts
mrmicahcooper: Hey @muenzpraeger the `lwc` packaged is checked on line 95.

mrmicahcooper:
@muenzpraeger Thanks for the awesome feedback. It should be addressed now:
- Fixed typos
- Fixed wrong description
- Now checking for `lwc` dependency and devDependency
- No longer checking for `lwc-services`

packages/lightning-lsp-common/src/shared.ts
mysticflute: We have a story to update the LWC version in this project but unfortunately it will be some work because a lot of the types have been removed to the platform compiler or removed.

packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
```diff
@@ -30,6 +30,18 @@ describe('detectWorkspaceType', () => {
         expect(workspaceType).toEqual(WorkspaceType.SFDX);
     });
 
+    test('when an lwc.config.js file is present, workspaceType is STANDARD_LWC', () => {
+        mockFs({
+            workspacedir: {
+                'lwc.config.js': '',
```
muenzpraeger, line 36: @xyc This is not correct, as it's `lwc.config.json`, see [here](https://github.com/forcedotcom/lightning-language-server/pull/16/files#diff-fbeef00ae19df8f84ca0d4a7c8cee648R33). See also [here](https://rfcs.lwc.dev/rfcs/lwc/0020-module-resolution#lwc-configuration) in the docs.

packages/lightning-lsp-common/src/shared.ts
```diff
@@ -73,25 +73,39 @@ export function detectWorkspaceHelper(root: string): WorkspaceType {
         return WorkspaceType.CORE_PARTIAL;
     }
 
+    if (fs.existsSync(path.join(root, 'lwc.config.js'))) {
```
rsalvador, line 76: @xyc @mrmicahcooper here it should also have been 'lwc.config.json'

xyc: Thanks we'll take a look. @muenzpraeger also mentions it

packages/lightning-lsp-common/src/__tests__/workspace-type.test.ts
mrmicahcooper: Nice catch. 
Here's a fix for this:
https://github.com/forcedotcom/lightning-language-server/pull/152/files

Thank you!

packages/lightning-lsp-common/src/shared.ts
muenzpraeger: It's already fixed in #16 (sorry ;-)

PR #124, opened 2020-05-18
Save Static Resource Index to file and initialize with it [Proof of concept]

mrmicahcooper:
Pre-requisite: Merge https://github.com/forcedotcom/lightning-language-server/pull/123 as it this builds on top of it.

### What does this PR do?
In an effort to prevent the indexer from duplicating effort across Aura and LWC language servers:
- When the language server shuts down, save its `STATIC_RESOURCES` index as a JSON file located in `./.sfdx/indexes/lwc/staticresources.json`
- When the language server indexes static components for the first time, load from a said file instead of regenerating the entire index.
- This should only occur if the in-memory static resources index is empty.


mrmicahcooper:
Moving this PR to a more appropriately named branch. Closing.

PR #136, opened 2020-05-27
Introduce save and load static resources from file

mrmicahcooper:
### What does this PR do?
Lay the foundation for loading all indexes from a file on server initialization

This PR does not change functionality. I simply introduces the methods which will be used in a later PR.

### DIFF
--- a/packages/lwc-language-server/src/metadata-utils/static-resources-util.ts
+++ b/packages/lwc-language-server/src/metadata-utils/static-resources-util.ts
@@ -8,7 +8,8 @@ import * as fs from 'fs-extra';
 const glob = promisify(Glob);
 
 const STATIC_RESOURCE_DECLARATION_FILE = '.sfdx/typings/lwc/staticresources.d.ts';
-const STATIC_RESOURCES: Set<string> = new Set();
+const STATIC_RESOURCE_INDEX_FILE = '.sfdx/indexes/lwc/staticresources.json';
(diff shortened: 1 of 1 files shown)


packages/lwc-language-server/src/metadata-utils/static-resources-util.ts
```diff
@@ -19,60 +20,91 @@ function getResourceName(resourceMetaFile: string) {
...
-    if (didChange) {
-        return processStaticResources(workspaceRoots[0], writeConfigs);
+        if (didChange) {
+            return processStaticResources(workspaceRoots[0], writeConfigs);
+        }
     }
 }
 
 async function processStaticResources(workspace: string, writeConfigs: boolean): Promise<void> {
     if (STATIC_RESOURCES.size > 0 && writeConfigs) {
```
rcoringrato-sfdc, line 46: @mrmicahcooper Is this logic accurate for the case where STATIC_RESOURCES size changes from 1 to 0?

mrmicahcooper: Hmm we didn't change that logic. This is how it was previously written. That being said, if you did delete all of your static resources, it does seem like this would not trigger for the last remaining resource.

I'll make a note to revisit if that's ok.

rcoringrato-sfdc: @mrmicahcooper Yes, that sounds good, thanks.

PR #160, opened 2020-07-01
Re-implement LWC server

mrmicahcooper:


packages/lwc-language-server/src/resources/convert.js
```diff
@@ -0,0 +1,82 @@
...
+
+    if (attribute.name) {
+        return '* **' + attribute.name + '**';
+    }
+
+    return '';
+}
+
+// read old file
+const f = fs.readFileSync('lwc-standard.json', { encoding: 'utf-8' });
```
lcampos, line 35: `lwc-standard.json` will keep growing as time goes by. Is it worth switching to nodejs streams for read and write ?

packages/lwc-language-server/src/base-indexer.ts
```diff
@@ -0,0 +1,21 @@
...
+    readonly workspaceRoot: string;
+    constructor(attributes: { workspaceRoot: string }) {
+        this.workspaceRoot = path.resolve(attributes.workspaceRoot);
+    }
+
+    sfdxConfig(root: string) {
+        const filename: string = path.join(root, 'sfdx-project.json');
+        const data: string = fsExtra.readFileSync(filename).toString();
+
+        return JSON.parse(data);
```
lcampos, line 13: `sfdx-project.json` is a file customers might change often and we've seen cases where they have invalid json. We should add some error handling for this scenario and throw an error message that users can act on.

packages/lwc-language-server/src/component-indexer.ts
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    get customData(): Tag[] {
+        return Array.from(this.tags.values());
+    }
+
+    findTagByName(query: string): Tag | null {
+        const matches = componentPrefixRegex.exec(query);
+        const { type, name } = matches.groups;
```
rcoringrato-sfdc, line 69: This should be guarded in case of null matches. It produces an error for certain onDefinition() calls.

PR #181, opened 2020-07-28
Update publish circleci flow

lcampos:
### What does this PR do?
Update publish flow so it doesn't commit when a release fails.

@W-7785612@

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -104,11 +104,24 @@ jobs:
       - add_ssh_keys: *ssh-config
       - checkout
       - run: *gh-config
-      - run: npm config set '//registry.npmjs.org/:_authToken' "${NPM_PUBLISH_TOKEN}"
-      - run: echo "Bumping version '<< pipeline.parameters.version >>''"
-      - run: |
-          yarn bump-versions << pipeline.parameters.version >>
-          yarn publish-lsp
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -104,11 +104,24 @@ jobs:
...
+          command: |
+            yarn bump-versions << pipeline.parameters.version >>
+            git add .
+            export RELEASE_TAG="$(node -pe "require('./packages/lightning-lsp-common/package.json').version")"
+            git commit -m "Updated version $RELEASE_TAG"
+      - run:
+          name: Set npm token
+          command: npm config set '//registry.npmjs.org/:_authToken' "${NPM_PUBLISH_TOKEN}"
+      - run: yarn publish-lsp
+      - run: git push origin main
```
xyc, line 118: move this to the next step too?

sfsholden: We need this step after the publish in the event that publish fails. We recently tried to publish and it failed, but pushed these changes into main anyways. This will prevent that going forward.

xyc: I'm referring to this line: `git push origin main`

xyc: and merge this with the next step. would that not prevent pushing changes after publishing fails?

PR #189, opened 2020-08-15
LWC bug fixes

lcampos:
### What does this PR do?
This PR fixes the following issues:
- Receiving events without data to process
- Persisting custom component information when the file path does not exist
- Persisting custom component information after indexing a project
- Not showing correct custom component information when triggering onCompletion for javascript lwc files

@W-7971496@

### DIFF
--- a/packages/lwc-language-server/src/component-indexer.ts
+++ b/packages/lwc-language-server/src/component-indexer.ts
@@ -4,13 +4,14 @@ import { shared } from '@salesforce/lightning-lsp-common';
 import { Entry, sync } from 'fast-glob';
 import normalize from 'normalize-path';
 import * as fsExtra from 'fs-extra';
-import { join } from 'path';
+import * as fs from 'fs';
 import { snakeCase } from 'change-case';
 import camelcase from 'camelcase';
 import BaseIndexer from './base-indexer';
 
 const { detectWorkspaceHelper, WorkspaceType } = shared;
-const CUSTOM_COMPONENT_INDEX_FILE = path.join('.sfdx', 'indexes', 'lwc', 'custom-components.json');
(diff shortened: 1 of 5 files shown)


packages/lwc-language-server/src/component-indexer.ts
```diff
@@ -103,7 +112,8 @@ export default class ComponentIndexer extends BaseIndexer {
     }
 
     persistCustomComponents() {
-        const indexPath = join(this.workspaceRoot, CUSTOM_COMPONENT_INDEX_FILE);
+        const indexPath = path.join(this.workspaceRoot, CUSTOM_COMPONENT_INDEX_FILE);
```
lcampos, line 115: This was failing on projects that didn't previously had `.sfdx/indexes/lwc/` created.

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -118,6 +127,18 @@ export default class Server {
         if (await this.context.isLWCTemplate(doc)) {
             this.auraDataProvider.activated = false; // provide completions for lwc components in an Aura template
             this.lwcDataProvider.activated = true;
+        } else if (await this.context.isLWCJavascript(doc)) {
```
lcampos, line 130: If the completion event is triggered from a lwc js file we get the already cached tag info and add their `c/<component name>` info to the rest of the suggestions provided by the html lang server.

packages/lwc-language-server/src/tag.ts
```diff
@@ -32,7 +32,7 @@ export default class Tag implements ITagData {
         this.updatedAt = attributes.updatedAt ? new Date(attributes.updatedAt) : null;
```
rcoringrato-sfdc, line 32: This assignment seems redundant.

lcampos: Yeah, let me remove that line

PR #200, opened 2020-08-20
Missed dependencies

rorolopetegui:
This PR adds dependencies that were missed and were required in the code.

The Ticket corresponding to this changes is: W-7976730


lcampos:
@rorolopetegui the builds are currently failing because there are still multiple lint errors on `aura-language-server` and `lwc-language-server`. Let's scope this PR down to exclusively updating `lightning-lsp-common` to eslint and we can address the other two packages in separate PRs.

rorolopetegui:
> @rorolopetegui the builds are currently failing because there are still multiple lint errors on `aura-language-server` and `lwc-language-server`. Let's scope this PR down to exclusively updating `lightning-lsp-common` to eslint and we can address the other two packages in separate PRs.

As we spoke earlier today. I reset this PR so we can sepparate things from linter and dependencies... In this PR were mixed both.
Now the only problem that we're having here, its the one that you tell me today @lcampos

PR #201, opened 2020-08-20
lsp-common/Different-Environment-os

rorolopetegui:
This PR adds the logic so when selecting the routes for different OS, you can run the tests in windows and other OS

The ticket associated with this PR is: W-7976730


rorolopetegui:
This PR is redundant, we got windowsCopySymlinks script so we can copy the workspaces for the tests.

PR #202, opened 2020-08-20
Bump of vscode-uri version

rorolopetegui:
This PR updates the dependency vscode-uri, also it changes the code where it uses it, cause URI is no longer exported as default.

The Ticket corresponding to this changes is: W-7976730

### DIFF
--- a/packages/lwc-language-server/package.json
+++ b/packages/lwc-language-server/package.json
@@ -46,7 +46,7 @@
     "vscode-html-languageservice": "^3.1.0-next.3",
     "vscode-languageserver": "^5.2.1",
     "vscode-nls": "^4.0.0",
-    "vscode-uri": "1.0.6",
+    "vscode-uri": "^2.1.2",
     "xml2js": "^0.4.19"
   },
   "devDependencies": {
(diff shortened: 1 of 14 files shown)


packages/lwc-language-server/package.json
```diff
@@ -46,7 +46,7 @@
     "vscode-html-languageservice": "^3.1.0-next.3",
     "vscode-languageserver": "^5.2.1",
     "vscode-nls": "^4.0.0",
-    "vscode-uri": "1.0.6",
+    "vscode-uri": "^2.1.2",
```
lcampos, line 49: Looks like there are more changes needed in this module since tests are failing with
```
@salesforce/lwc-language-server: src/__tests__/test-utils.ts:14:21 - error TS2339: Property 'file' does not exist on type 'typeof import("C:/Users/circleci/project/node_modules/vscode-uri/lib/umd/index")'.
@salesforce/lwc-language-server: 14     const uri = URI.file(resolve(path)).toString();
```

rorolopetegui: Nice Catch here!
I fix this, and run all tests now: 
image

PR #203, opened 2020-08-20
Bump dependencies

rorolopetegui:
This PR bumps some dependencies that dont needed an update on the code

The Ticket corresponding to this changes is: W-8001765


packages/aura-language-server/package.json
```diff
@@ -31,38 +31,39 @@
...
-    "tsc-transpile-only": "^0.0.3",
+    "prettier": "^2.0.5",
+    "shelljs": "^0.8.4",
+    "ts-jest": "^26.2.0",
+    "ts-node": "^8.10.2",
+    "tsc-transpile-only": "0.0.3",
     "tslint": "^6.0.0",
     "tslint-config-prettier": "^1.18.0",
-    "typescript": "^3.1.6"
+    "typescript": "^4.0.2"
```
lcampos, line 67: Is there any breaking change between 3 & 4 ? These lang servers are used by the vscode extension which are currently on typescript 3.7.5 (https://github.com/forcedotcom/salesforcedx-vscode/blob/810aba681b2cfabf55a9019ff7f4e6910f57ced1/packages/salesforcedx-vscode-lightning/package.json#L50) and it's important we use compatible versions of typescript.

packages/aura-language-server/package.json
```diff
@@ -31,38 +31,39 @@
     "acorn-loose": "^6.0.0",
     "acorn-walk": "^6.0.0",
     "change-case": "^3.1.0",
-    "enhanced-resolve": "^2.2.2",
-    "fs-extra": "^5.0.0",
+    "enhanced-resolve": "^4.3.0",
+    "fs-extra": "^9.0.1",
```
lcampos, line 35: What about similar updates to `lightning-lsp-common` ? Some of the modules here are also used in that package.

packages/aura-language-server/package.json
xyc: TypeScript doesn't adhere to strict sem versioning - but agreeing to @lcampos's point we should validate TS versions are compatible

packages/aura-language-server/package.json
rorolopetegui: Luis, as we spoke yesterday, now all the versions are corresponding

packages/aura-language-server/package.json
rorolopetegui: As we spoke yesterday with Luis, we said that we were let the 3.7.5 version into this packages

packages/aura-language-server/package-lock.json
```diff
@@ -0,0 +1,10263 @@
+{
```
lcampos, line 1: We build this project using `yarn` so let's remove this lock file which is specific to `npm`

packages/lwc-language-server/package.json
```diff
@@ -36,38 +35,39 @@
...
-    "xml2js": "^0.4.19"
+    "xml2js": "^0.4.23"
   },
   "devDependencies": {
-    "@types/babel-types": "^7.0.4",
-    "@types/fs-extra": "^5.0.4",
-    "@types/glob": "^5.0.35",
-    "@types/jest": "^25.1.2",
-    "@types/node": "^10.12.18",
+    "@lwc/engine": "0.34.8",
```
lcampos, line 52: Are we sure we can more `@lwc/engine` to be on `devDependencies` ?

rorolopetegui: We are using this @lwc/engine on 2 places, one is a test called workspace-type.test.ts, and the other is a script called copy_typings.js that's used in the build script. Thats what i found.
Also, we're using this dependency in the other packages as Dev dependency. 
I think we should let this dependency on the devs

packages/lightning-lsp-common/package.json
```diff
@@ -44,27 +43,28 @@
...
     "@types/lodash": "^4.14.119",
     "@types/lodash.template": "^4.4.4",
     "@types/mock-fs": "^4.10.0",
     "@types/semver": "^5.5.0",
     "@types/tmp": "^0.1.0",
     "find-node-modules": "^1.0.4",
     "findup-sync": "^3.0.0",
-    "husky": "^4.2.3",
-    "jest": "^25.1.0",
+    "glob": "^7.1.6",
```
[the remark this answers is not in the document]
rorolopetegui: You're right, i thought that i was in another package, we're using this library in a test (in the other packages).

packages/lwc-language-server/package.json
rorolopetegui: Its strange, because this dependency is not imported anywhere for lwc-language-server... Its just used by the other ones for some test.
Maybe its imported but not used, i'll let the @lwc/engine onto the dependencies, and then make a check of unused dependencies (I saw that was one of the pending tasks to do)

rorolopetegui: Also when removed, no error is showed on the tests

lcampos: We unfortunately can't fully rely on the tests not failing to show a problem. Let's figure out whether we need this dependency or not for a future PR.

lcampos:
@rorolopetegui the changes look good, we just need to figure out why CircleCI is failing 
```
@salesforce/lightning-lsp-common: $ tsc -p . && yarn copy-static-assets && node ./scripts/copy_typings.js && cp -r src/resources lib/
@salesforce/lightning-lsp-common: ../../node_modules/@types/fs-extra/index.d.ts:188:87 - error TS2694: Namespace '"fs"' has no exported member 'Dir'.
@salesforce/lightning-lsp-common: 188 export function opendir(path: string, cb: (err: NodeJS.ErrnoException | null, dir: fs.Dir) => void): void;
```

PR #207, opened 2020-08-31
Missed package @salesforce/apex

rorolopetegui:
This PR adds the dependency @salesforce/apex into the lightning-lsp-common package. 
It's copied on scripts/copy_typings.js, and used on test lightning-lsp-common/src/_tests_/context.test.ts:201 (Where it checks for apex.d.ts file)

### DIFF
--- a/packages/lightning-lsp-common/package.json
+++ b/packages/lightning-lsp-common/package.json
@@ -27,7 +27,6 @@
   },
   "dependencies": {
     "@lwc/babel-plugin-component": "0.34.8",
-    "@salesforce/apex": "0.0.12",
     "decamelize": "^2.0.0",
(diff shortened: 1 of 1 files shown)


xyc:
If this is only used in `scripts/copy_typings.js` and tests I think it could be a devDependency, since the result npm package wouldn't require installing `@salesforce/apex`. @rorolopetegui @lcampos thoughts?

lcampos:
> If this is only used in `scripts/copy_typings.js` and tests I think it could be a devDependency, since the result npm package wouldn't require installing `@salesforce/apex`. @rorolopetegui @lcampos thoughts?

Yes, let's do that @rorolopetegui

PR #208, opened 2020-09-01
Eslint fixes

rorolopetegui:
This PR uses eslint rules and fixes the errors throwed by the library
The configuration that we're using here is the same as the main repositories that we use.

The ticket related to this PR: W-8039500


packages/lightning-lsp-common/src/__tests__/context.test.ts
```diff
@@ -205,6 +204,48 @@ it('configureSfdxProject()', async () => {
...
+function verifyCoreSettings(settings: any) {
+    expect(settings['files.watcherExclude']).toBeDefined();
+    expect(settings['eslint.nodePath']).toBeDefined();
+    expect(settings['perforce.client']).toBe('username-localhost-blt');
+    expect(settings['perforce.user']).toBe('username');
+    expect(settings['perforce.port']).toBe('ssl:host:port');
+}
+
+/*
+function verifyCodeWorkspace(path: string) {
```
lcampos, line 234: Is this not needed anymore ?

packages/lightning-lsp-common/src/utils.ts
```diff
@@ -13,27 +13,68 @@ export const glob = promisify(Glob);
...
+}
+
 /**
  * @return true if changes include a directory delete
  */
+// TODO This is not waiting for the response of the promise isLWCWatchedDirectory, maybe we have the same problem on includesDeletedAuraWatchedDirectory
 export function includesDeletedLwcWatchedDirectory(context: WorkspaceContext, changes: FileEvent[]): boolean {
     for (const event of changes) {
-        if (event.type === FileChangeType.Deleted && isLWCWatchedDirectory(context, event.uri)) {
+        if (event.type === FileChangeType.Deleted && event.uri.indexOf('.') === -1 && isLWCWatchedDirectory(context, event.uri)) {
```
lcampos, line 62: This should await on `isLWCWatchedDirectory` and will have to add `async` to the method signature.

packages/lightning-lsp-common/src/utils.ts
```diff
@@ -13,27 +13,68 @@ export const glob = promisify(Glob);
...
             return true;
         }
     }
     return false;
 }
 export function includesDeletedAuraWatchedDirectory(context: WorkspaceContext, changes: FileEvent[]): boolean {
     for (const event of changes) {
-        if (event.type === FileChangeType.Deleted && isAuraWatchedDirectory(context, event.uri)) {
+        if (event.type === FileChangeType.Deleted && event.uri.indexOf('.') === -1 && isAuraWatchedDirectory(context, event.uri)) {
```
lcampos, line 70: Similar changes as needed in `includesDeletedLwcWatchedDirectory`

yarn.lock
```diff
@@ -1534,26 +1388,23 @@
 "@lerna/timer@3.13.0":
   version "3.13.0"
   resolved "https://registry.yarnpkg.com/@lerna/timer/-/timer-3.13.0.tgz#bcd0904551db16e08364d6c18e5e2160fc870781"
-  integrity sha512-RHWrDl8U4XNPqY5MQHkToWS9jHPnkLZEt5VD+uunCKTfzlxGnRCr3/zVr8VGy/uENMYpVP3wJa4RKGY6M0vkRw==
 
 "@lerna/validation-error@3.13.0":
   version "3.13.0"
   resolved "https://registry.yarnpkg.com/@lerna/validation-error/-/validation-error-3.13.0.tgz#c86b8f07c5ab9539f775bd8a54976e926f3759c3"
-  integrity sha512-SiJP75nwB8GhgwLKQfdkSnDufAaCbkZWJqEDlKOUPUvVOplRGnfL+BPQZH5nvq2BYSRXsksXWZ4UHVnQZI/HYA==
```
lcampos, line 1542: Why is all the integrity data being removed ?

.huskyrc
```diff
@@ -1,6 +0,0 @@
-{
```
lcampos, line 1: It's ok to delete the .huskyrc configs since we want to have them align with what we're doing in the other repos 👍

rorolopetegui: Ok, nice! Thanks!

yarn.lock
rorolopetegui: My bad, i deleted the yarn.lock on a test that i was doing... I will restore this file

packages/lightning-lsp-common/src/__tests__/context.test.ts
rorolopetegui: I think we should remove this function, because from where it is called, it is commented. Or maybe let the function commented, so in a future if we need it, we can modify it.

.huskyrc
rorolopetegui: Ok, i was looking and when i pushed the other changes, it was not passing the tests on CircleCi, it was because its running yarn with lock files (I think thats because when you yarn install, that can take some time).
So i re-upload the yarn.lock again

yarn.lock
rorolopetegui: Ok, i was looking and when i pushed the other changes, it was not passing the tests on CircleCi, it was because its running yarn with lock files (I think thats because when you yarn install, that can take some time).
So i re-upload the yarn.lock again

lcampos (approved):

PR #216, opened 2020-09-09
Move script build to node script

rorolopetegui:
### What does this PR do?
This PR tries to implement one way of builind each package, to consolidate a way of doing it.

W-7976902


packages/aura-language-server/scripts/build.js
```diff
@@ -0,0 +1,8 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
lcampos, line 1: This script should contain the logic in `copy-static-assets.ts` so we can delete that ts file and not have to run it using ts-node by using `copy-static-assets` script in package.json

packages/lightning-lsp-common/scripts/build.js
```diff
@@ -0,0 +1,17 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const shell = require('shelljs');
+
+// Copy static assets
+if (shell.exec('yarn copy-static-assets"').code !== 0) {
```
lcampos, line 5: Same as the comments for the script in aura-language-server, let's move the logic in `copy-static-assets.ts` here so we can remove the use of ts-node and that ts file

rorolopetegui: Yeah, i was about to do that, as ts-node is only used to run that single file

packages/aura-language-server/scripts/build.js
```diff
@@ -1,5 +1,7 @@
-import * as shell from 'shelljs';
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
lcampos, line 1:
```suggestion
#!/usr/bin/env node
// eslint-disable-next-line @typescript-eslint/no-var-requires
```

packages/lightning-lsp-common/scripts/build.js
```diff
@@ -0,0 +1,15 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
lcampos, line 1:
```suggestion
#!/usr/bin/env node
// eslint-disable-next-line @typescript-eslint/no-var-requires
```

packages/lightning-lsp-common/scripts/copy_typings.js
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env node
+/* eslint-disable @typescript-eslint/no-var-requires */
```
lcampos, line 1:
```suggestion
#!/usr/bin/env node
/* eslint-disable @typescript-eslint/no-var-requires */
```

packages/lwc-language-server/scripts/build.js
```diff
@@ -0,0 +1,5 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
```
lcampos, line 1:
```suggestion
#!/usr/bin/env node
// eslint-disable-next-line @typescript-eslint/no-var-requires
```

lcampos (changes requested):
Other than the suggestions to add the shebang, we need to change the script's permissions to be executable. We used `chmod 755 <file path>` on the vscode repo scripts

packages/aura-language-server/package.json
```diff
@@ -12,9 +12,8 @@
   "scripts": {
     "clean": "rm -rf lib",
     "cleanTestData": "node ../../scripts/removeTestData.js",
-    "build": "tsc -p . && yarn copy-static-assets",
+    "build": "tsc -p . && node scripts/build.js",
```
xyc, line 15:
```suggestion
    "build": "tsc -p . && scripts/build.js",
```
We don't need to specify node with @lcampos 's shebang suggestions (and adding execute permission to the script).

rorolopetegui: I cant take out the node command, because on windows its needed. If i remove it, i will prompt an error with the windows script host. The CircleCI test for windows will never pass, and it will show a "timeout error"(I think because of this error).
This is the error: 
image

packages/lightning-lsp-common/scripts/copy_typings.js
```diff
@@ -1,7 +1,7 @@
-#!/usr/bin/env node
```
lcampos, line 1: Let's keep the node shebang

lcampos (approved):

packages/aura-language-server/package.json
xyc: Sounds good, yeah windows doesn't support shebang and it's interpreted as JScript. Let's keep as is then.

xyc (approved):

PR #228, opened 2020-09-16
Rlopetegui/update lwc compiler dependency

rorolopetegui:
### What does this PR do?
This PR updates the @lwc/compiler dependency, and changes the code for that was using this dependency functions

W-8085326


rorolopetegui:
This PR is blocked for now as we cant update this dependency cause the metadata that we use, were removed from the compiler. We're in touch with the LWC Framework team, trying to get into a solution.

PR #243, opened 2020-09-30
Trigger completions on specific characters

lcampos:
### What does this PR do?
Update completion provider configs to trigger the event when a user types aura/lwc tag characters.

@W-8017548@, @W-8017525@

#### Aura Before
aura-completions-before

#### Aura After
aura-completions-after

#### LWC Before
lwc-completions-before

#### LWC After
lwc-completions-after

### DIFF
--- a/packages/lwc-language-server/src/lwc-server.ts
+++ b/packages/lwc-language-server/src/lwc-server.ts
@@ -116,7 +116,7 @@ export default class Server {
                 textDocumentSync: this.documents.syncKind,
                 completionProvider: {
                     resolveProvider: true,
-                    triggerCharacters: ['-', '_', '<', '"', '=', '/', '>'],
+                    triggerCharacters: ['{', '.', '-', '_', '<', '"', '=', '/', '>'],
                 },
                 hoverProvider: true,
                 definitionProvider: true,
--- a/packages/lwc-language-server/src/lwc-server.ts
+++ b/packages/lwc-language-server/src/lwc-server.ts
@@ -116,6 +116,7 @@ export default class Server {
                 textDocumentSync: this.documents.syncKind,
                 completionProvider: {
                     resolveProvider: true,
+                    triggerCharacters: ['-', '_', '<', '"', '=', '/', '>'],
                 },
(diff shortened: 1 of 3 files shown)


packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -116,6 +116,7 @@ export default class Server {
                 textDocumentSync: this.documents.syncKind,
                 completionProvider: {
                     resolveProvider: true,
+                    triggerCharacters: ['-', '_', '<', '"', '=', '/', '>'],
```
xyc, line 119: should `'.'` be included too? https://microsoft.github.io/language-server-protocol/specification#textDocument_completion

xyc (approved):

packages/lwc-language-server/src/lwc-server.ts
lcampos: That's a good question, I initially had it but I couldn't find an example to justify having that character around. In which situations would we have aura or lwc refs after a `.` character ?

xyc: I mentioned it since the above doc uses `.` as an example.
> If code complete should automatically be trigger on characters not being valid inside an identifier (for example `.` in JavaScript) list them in `triggerCharacters`.

> In which situations would we have aura or lwc refs after a . character ?

For example data binding in html templates`{data.name}` https://lwc.dev/guide/html_templates#data-binding

sfsholden (approved):
LGTM - if it's possible, I think we'd benefit from some auto-completion test examples.

packages/lwc-language-server/src/lwc-server.ts
lcampos: Yeah, that requires a different change which I'll take care of with W-8017495 but I've added those characters for now since it could also help in some js scenarios.

lcampos:
> LGTM - if it's possible, I think we'd benefit from some auto-completion test examples.

This should cover any type of reference between aura components and lwc components that the frameworks support e.g. parent-child reference, etc

PR #263, opened 2020-10-13
chore: update port logic to handle errors and updates for local branches

sfsholden:
### What does this PR do?
1. Adds checks for error codes instead of strings. If an error code is not 0, we will exit with an error.
2. Changes the local branch update logic. Previously this was not accounting for fetches being performed on the current branch. By default this isn't allowed, which was causing an error. We can make use of the '-u' option, but per git man I don't believe we should use this flag. Instead, we will now just checkout main and do a pull to avoid this issue. It also ensures we always have the latest version as well.

-u flag Information:
> By default git fetch refuses to update the head which corresponds to the current branch. This flag disables the check. This is purely for the internal use for git pull to communicate with git fetch, and unless you are implementing your own Porcelain you are not supposed to use it.

@W-8189197@

### DIFF
--- a/scripts/port-changes.js
+++ b/scripts/port-changes.js
@@ -7,165 +7,216 @@
  * to the new branch.
  */
 
-const shell = require('shelljs');
+const shell = require("shelljs");
 
 const PR_REGEX = new RegExp(/(\(#\d+\))(\s+\(#\d+\))*$/);
 const COMMIT_REGEX = new RegExp(/^([\da-zA-Z]+)/);
 
-const PR_NUM = 'PR_NUM';
-const COMMIT = 'COMMIT';
-const MESSAGE = 'MESSAGE';
+const PR_NUM = "PR_NUM";
+const COMMIT = "COMMIT";
+const MESSAGE = "MESSAGE";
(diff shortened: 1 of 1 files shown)


scripts/port-changes.js
```diff
@@ -7,165 +7,216 @@
...
-const shell = require('shelljs');
+const shell = require("shelljs");
 
 const PR_REGEX = new RegExp(/(\(#\d+\))(\s+\(#\d+\))*$/);
 const COMMIT_REGEX = new RegExp(/^([\da-zA-Z]+)/);
 
-const PR_NUM = 'PR_NUM';
-const COMMIT = 'COMMIT';
-const MESSAGE = 'MESSAGE';
+const PR_NUM = "PR_NUM";
```
lcampos, line 15: was there a change on the eslint rules ?

lcampos (approved):

scripts/port-changes.js
sfsholden: It didn't look like it, this is what the formatter gives me now though. We can sync about this after standup.

PR #369, opened 2021-04-30
feat(common): add analyticsWaveApi types

smithgp:
### What does this PR do?
Adds typings for lightning/analyticsWaveApi lds methods.

@W-8973196@

cc: @mblumreich, @ychhetri, @lindsey-burnett, @danny6514, @yijian-wang, @reidaelliott, @karihotchkiss

### DIFF
--- a/packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
+++ b/packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
@@ -547,3 +547,814 @@ declare module 'lightning/uiRecordApi' {
      */
     export function getFieldDisplayValue(record: RecordRepresentation, field: FieldId | string): FieldValueRepresentationValue | undefined;
 }
+
+declare module 'lightning/analyticsWaveApi' {
+    /**
+     * A Wave dataflow node.
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface AbstractDataflowNodeRepresentation {
+        /** Node action */
+        action: string;
+        /** Node sources */
+        sources: Array<string>;
+    }
+
+    /**
+     * Asset reference representation.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_resources_appendix.htm#AssetReferenceRepresentation
+     *
+     * Keys:
+     *    (none)
+     */
+    // eslint-disable-next-line @typescript-eslint/no-empty-interface
+    export interface AssetReferenceRepresentation extends BaseAssetReferenceRepresentation {}
+
+    /**
+     * Base asset reference representation.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_resources_appendix.htm#AssetReferenceRepresentation
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface BaseAssetReferenceRepresentation {
+        /** The 18 character ID of the asset. */
+        id: string;
+        /** The asset label. */
+        label?: string;
+        /** The asset developer name. */
+        name?: string;
+        /** The namespace that qualifies the asset name */
+        namespace?: string;
+        /** The asset URL. */
+        url?: string;
+    }
+
+    /**
+     * Base Wave asset representation.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_resources_appendix.htm#BaseWaveAssetRepresentation
+     * Keys:
+     *    id (string): id
+     */
+    export interface BaseWaveAssetRepresentation {
+        /** Sharing URL for the asset. */
+        assetSharingUrl?: string | null;
+        /** The user that created the asset. */
+        createdBy?: WaveUserRepresentation;
+        /** Time the asset was created. */
+        createdDate?: string;
+        /** Short description of the asset. */
+        description?: string;
+        /** The 18 character asset ID. */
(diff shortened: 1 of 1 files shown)


packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -547,3 +547,814 @@ declare module 'lightning/uiRecordApi' {
...
 }
+
+declare module 'lightning/analyticsWaveApi' {
+    /**
+     * A Wave dataflow node.
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface AbstractDataflowNodeRepresentation {
```
smithgp, line 558: Most of the interfaces and their jsdoc is from the generated types in lds.

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -547,3 +547,814 @@ declare module 'lightning/uiRecordApi' {
...
+    /**
+     * Daily schedule on which to execute some type of job.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_resources_appendix.htm#DailyScheduleInputRepresentation
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface DailyScheduleInputRepresentation extends ScheduleInputRepresentation {
+        frequency: 'daily' | 'Daily';
```
smithgp, line 644: This is a connect enum used as a discriminator in lds, which comes through as a discriminated union type. Wave connect enums are case-insensitive on input and generally come through as camel-case in output, hence this `|`ed type on the various `frequency`'s.

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -547,3 +547,814 @@ declare module 'lightning/uiRecordApi' {
...
+    export interface WeeklyScheduleRepresentation extends ScheduleRepresentation {
+        frequency: 'weekly' | 'Weekly';
+        /** Days of the week on which the schedule will run. */
+        daysOfWeek: Array<string>;
+    }
+
+    /**
+     * Creates a Tableau CRM dataflow job, which is the equivalent of clicking Run Now for a data prep recipe, a data sync,
+     * or a dataflow in the Tableau CRM Data Manager UI.
+     *
```
smithgp, line 1121: The Developer Guide docs for these are currently in progress. Once their published and public, I'll update these methods here with links to them.

ychhetri (approved):

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
brpowell: I'm wondering if in an intellisense context when it shows both values it could be confusing whether there's a difference between capital or not. Does it make sense to have the docstring for the base interface's `frequency` prop to call out it is case-insensitive?

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -547,3 +547,814 @@ declare module 'lightning/uiRecordApi' {
...
+
+    /**
+     * Wire adapter to retrieve the Analytics limits for Tableau CRM.
+     *
+     * @param licenseType The response includes dataflow jobs with this license type. Valid values are `EinsteinAnalytics` or `Sonic`.
+     * @param types The types of limits used in Tableau CRM.
+     *              Valid values are `BatchTransformationHours`, `DatasetQueries`, `DatasetRowCount`,
+     *              `MaxDailyRowsHighOutputCon`, `MaxDailyRowsLowOutputCon`, `MaxDailyRowsMedOutputCon`,
+     *              `MaxDailySizeHighOutputCon`, `MaxDailySizeLowOutputCon`, `MaxDailySizeMedOutputCon`,
+     *              or `OutputLocalConnectorVolume`.
```
brpowell, line 1161: If you'd like, you could have type checking for the arguments using a type alias:

```typescript
type LicenseType = 'EinsteinAnalytics' | 'Sonic';
type LimitType = 'BatchTransformationHours' | 'DatasetQueries' | 'DatasetRowCount' // ...
export function getAnalyticsLimits(licenseType: LicenseType, types?: LimitType[]): void;
```

brpowell (approved):

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
smithgp: Makes sense, done.

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
smithgp: So I didn't do that originally for a few reasons:
1. I saw that another enum-ish field (`FieldRepresentation.extraType`) just did `string`.
2. This function can only be used with `@wire`, and therefore is allowed to be bound ala `@wire(getAnalyticsLimits, { licenseType: '$someVariable' })`, which, in type terms, is just a `string`.
3. The typedef for `@wire` currently has the 2nd arg (which corresponds to an object structure of the args for the 1st arg method) as `any`, so you don't get any completions or type checks there anyways.

I think I'm going to leave this as-is for now, until if/when there's better intellisense/type-checking around the 2nd arg of `@wire`.

brpowell: Makes sense, thanks for the additional context @smithgp

PR #379, opened 2021-05-10
fix: fix issue with automation token

xyc:
### What does this PR do?
We have issues with publishing with lerna with npm automation token.
https://github.com/lerna/lerna/issues/2788


brpowell:
Do we know how this was working before?

xyc:
^ the above comments come from changing the base branch

brpowell (approved):

PR #454, opened 2021-10-04
fix: renamed master to main

jeffb-sfdc:
### What does this PR do?
Updates some of the occurrences of "master" to "main'


randi274 (approved):
Great change! Thanks for getting this turned around so quickly 👏

PR #468, opened 2022-01-13
fix(launch.json and package.json): all tests and build step

dbreese:
1. package.json updated to filter file first before copying it. 
2. launch.json updated to use lerna to run tests instead of jest directly which was failing.
3. README updates.
4. add missing dependency for jest types for vscode intellisense.

### What does this PR do?
1. Fixes issue with build when making string changes to lwc-standard.json. The build scripts were copying the files in src/resources to lib/resources AND THEN processing them from src/resources, so it would take 2 builds to get the files to actually appear in the correct spot. Now it creates standard-lwc.json in src/ and then copies it to lib. Ideally, it seems this shouldn't even be processed in the src/ folder, but that's for another PR later if needed.
2. Running `Launch All LWC Tests` from VSCode was failing, but it works from command line when running "yarn tests" or via lerna. I updated the launch config to run via lerna which is what yarn tests does (see package.json):
3. README updates to be more clear on ways to open workspace.
4. add missing dependency for jest types for vscode intellisense.
```
    "test": "lerna run test --stream --no-bail -- --colors",
```

1. Build issues.
2. Running tests.

@maliroteh-sf @jag-sfdc


README.md
```diff
@@ -48,16 +48,21 @@ npm run compile
...
-code ./vscode-workspaces/multiroot-simple.code-workspace
-```
-
-OR
-
-```
-From the lightning-language-server project, open the file 'multiroot-simple.code-workspace'
-Click button 'Open Workspace'
-```
+```
```
dbreese, line 51: Attempt to make this more clear and why there are two ways to open the workspace.

packages/lwc-language-server/package.json
```diff
@@ -11,7 +11,7 @@
   "scripts": {
     "clean": "rm -rf lib && rm -rf src/resources/sfdx/typings/copied",
     "cleanTestData": "node ../../scripts/removeTestData.js",
-    "build": "tsc -p . && node scripts/build.js && node scripts/convert.js",
+    "build": "tsc -p . && node scripts/convert.js && node scripts/build.js",
```
dbreese, line 14: This just flip-flops the steps. Without doing this, making any changes to src/lwc-standard.json takes two builds to eventually get to the lib directory.

dbreese:
Not sure why this is failing. Doesn't appear to be around anything I've touched: 

https://app.circleci.com/pipelines/github/forcedotcom/lightning-language-server/1885/workflows/7ff3464d-4a15-48be-882f-cb60c584b5f2/jobs/3582?invite=true#step-105-133

```
EAGAIN: resource temporarily unavailable, read
```

In test `unit src/aura-indexer/__tests__/indexer.spec.ts`

dbreese:
The broken test was commented out in commit 588223. https://github.com/forcedotcom/lightning-language-server/pull/472

randi274:
Added this in with #475 - thanks for the fix!

PR #481, opened 2022-02-16
chore: use context for common env variables

RitamAgrawal:
### What does this PR do?
Access common env variables across different projects from context defined at organizational settings.


gbockus-sf (approved):
👍

PR #482, opened 2022-02-16
Ra/context for env variables

RitamAgrawal:
### What does this PR do?
Uses context to access environment variables.


gbockus-sf (approved):
Good to go

PR #488, opened 2022-03-03
feat: add iterator syntax

randi274:
### What does this PR do?
Added the iterator:it template directive to suggested attributes with others like for:each and
if:true. Fixed issue where description wasn't showing on any of the directives, and added links to
our docs as well. This also allow for flexible syntax, where the user wants to use another variable 
for 'it'.

@W-10536569@, @W-10542295@

### What was the behavior before?
The template directive iterator:it was not appearing in our list of auto-complete attributes, and description information was broken.

### What was the behavior now?
The directive appears, and the description information is now available for all other template directives. 🎉

### DIFF
--- a/packages/lwc-language-server/scripts/convert.js
+++ b/packages/lwc-language-server/scripts/convert.js
@@ -50,26 +50,71 @@ const tags = Object.keys(data).map(key => {
     };
 });
 
+//make globalAttribute changes here, not in standard-lwc.json as they'll be overwritten
 const globalAttributes = [
     {
         name: 'for:each',
-        decription: 'Renders the element or template block multiple times based on the expression value.',
+        description: 'Renders the element or template block multiple times based on the expression value.',
+        references: [
+            {
+                name: 'Salesforce',
+                url: 'https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.reference_directives',
+            },
+        ],
     },
     {
         name: 'for:item',
-        decription: 'Bind the current iteration item to an identifier.',
+        description: 'Bind the current iteration item to an identifier.',
+        references: [
+            {
+                name: 'Salesforce',
+                url: 'https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.reference_directives',
+            },
+        ],
     },
     {
         name: 'for:index',
-        decription: 'Bind the current iteration index to an identifier.',
+        description: 'Bind the current iteration index to an identifier.',
+        references: [
+            {
+                name: 'Salesforce',
+                url: 'https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.reference_directives',
+            },
+        ],
     },
     {
         name: 'if:true',
-        decription: 'Renders the element or template if the expression value is thruthy.',
+        description: 'Renders the element or template if the expression value is truthy.',
+        references: [
+            {
+                name: 'Salesforce',
+                url: 'https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.reference_directives',
+            },
+        ],
     },
     {
         name: 'if:false',
-        decription: 'Renders the element or template if the expression value is falsy.',
+        description: 'Renders the element or template if the expression value is falsy.',
+        references: [
+            {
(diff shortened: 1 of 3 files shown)


packages/lwc-language-server/scripts/convert.js
```diff
@@ -50,26 +50,71 @@ const tags = Object.keys(data).map(key => {
     };
 });
 
+//make globalAttribute changes here, not in standard-lwc.json as they'll be overwritten
```
jeffb-sfdc, line 53: OK, this is super nin-picky, but I thought out ES-lint rules required a space after the `//`

jeffb-sfdc: Other than this, the PR looks good.

packages/lwc-language-server/src/resources/standard-lwc.json
```diff
@@ -3649,23 +3649,66 @@
   "globalAttributes": [
     {
       "name": "for:each",
-      "decription": "Renders the element or template block multiple times based on the expression value."
+      "description": "Renders the element or template block multiple times based on the expression value.",
```
gbockus-sf, line 3652: this typo found it's way everywhere :)

packages/lwc-language-server/scripts/convert.js
```diff
@@ -50,26 +50,71 @@ const tags = Object.keys(data).map(key => {
...
 const globalAttributes = [
```
gbockus-sf, line 54: nit: so what way we could have caught his is if 1. this was a ts file and 2. we defined a type/interface for the attributes to implement.

packages/lwc-language-server/scripts/convert.js
```diff
@@ -50,26 +50,71 @@ const tags = Object.keys(data).map(key => {
...
+            kind: 'markdown',
+            value:
+                'Bind the current iteration item to an identifier. Contains properties (`value`, `index`, `first`, `last`) that let you apply special behaviors to certain items.',
+        },
+        references: [
+            {
+                name: 'Salesforce',
+                url: 'https://developer.salesforce.com/docs/component-library/documentation/en/lwc/lwc.reference_directives',
+            },
+        ],
```
gbockus-sf, line 117: dumb q: are these dups for what's in [standard-lwc.json](https://github.com/forcedotcom/lightning-language-server/pull/488/files#diff-e30348f58226e23daf10a57ff87c0b636112f69a5de89aa0434c6344403d961d). If so what don't we just import that here and use it?

packages/lwc-language-server/scripts/convert.js
gbockus-sf: this is more just an observation.  no action to be taken here.

gbockus-sf (approved):
changes seem good.  One suggestion to consider if the json data and the js data are intended to be the same.

packages/lwc-language-server/scripts/convert.js
randi274: I think something's amiss with eslint here; @gbockus-sf gave similar feedback on a previous PR in here! I will get that done in short order though 👍

packages/lwc-language-server/src/resources/standard-lwc.json
randi274: It really did 😆

packages/lwc-language-server/scripts/convert.js
randi274: 💯

packages/lwc-language-server/scripts/convert.js
randi274: So we have a script that runs every time we build that calls `convert.js`, and that is what then populates `standard-lwc.json`. I was trying to dig up if there was still a purpose to having this separately, but wasn't able to find a clear answer in any conversation/PR/commit message. That said, globalAttributes goes back a long time in this file, and I do know for sure that this overwrites whatever we put in to standard-lwc.

gbockus-sf: gotcha

jeffb-sfdc (approved):

PR #490, opened 2022-03-15
fix: Prevent excessive autocompletion in template JS files

randi274:
### What does this PR do?
- Addresses a recent issue where we were providing autocomplete for curly braces in JS files, when we really only wanted to do so in HTML. This caused annoying and unhelpful autocomplete due to them firing off every time a conditional statement or method was declared. This reverts the JS autocomplete to how it was previously with a bit more semantic method naming to make it clear what the trigger characters are doing.

VS Code #3902. @W-10832653@

### DIFF
--- a/packages/lwc-language-server/src/__tests__/lwc-server.test.ts
+++ b/packages/lwc-language-server/src/__tests__/lwc-server.test.ts
@@ -10,6 +10,10 @@ const filename = path.resolve('../../test-workspaces/sfdx-workspace/force-app/ma
 const uri = URI.file(filename).toString();
 const document: TextDocument = TextDocument.create(uri, 'html', 0, fsExtra.readFileSync(filename).toString());
 
+const jsFilename = path.resolve('../../test-workspaces/sfdx-workspace/force-app/main/default/lwc/todo/todo.js');
+const jsUri = URI.file(jsFilename).toString();
+const jsDocument: TextDocument = TextDocument.create(uri, 'javascript', 0, fsExtra.readFileSync(jsFilename).toString());
+
 const auraFilename = path.resolve('../../test-workspaces/sfdx-workspace/force-app/main/default/aura/todoApp/todoApp.app');
 const auraUri = URI.file(auraFilename).toString();
 const auraDocument: TextDocument = TextDocument.create(auraFilename, 'html', 0, fsExtra.readFileSync(auraFilename).toString());
@@ -41,6 +45,7 @@ jest.mock('vscode-languageserver', () => {
                 get: (name: string): TextDocument => {
                     const docs = new Map([
                         [uri, document],
+                        [jsUri, jsDocument],
                         [auraUri, auraDocument],
                         [hoverUri, hoverDocument],
                     ]);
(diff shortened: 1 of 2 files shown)


packages/lwc-language-server/src/__tests__/lwc-server.test.ts
```diff
@@ -76,26 +81,44 @@ describe('handlers', () => {
...
             const params: CompletionParams = {
-                textDocument: { uri },
+                textDocument: { uri: jsUri },
                 position: {
                     line: 0,
                     character: 0,
                 },
                 context: {
-                    triggerCharacter: '{',
+                    triggerCharacter: '.',
```
randi274, line 90: This test covers the autocomplete that was already being performed and not tested. There's a list of [other](https://github.com/forcedotcom/lightning-language-server/blob/34a0257b8f0b6ddb9f8caf7a6e37741e329b81f9/packages/lwc-language-server/src/lwc-server.ts#L120) trigger characters that would also work here, some of which are also not terribly helpful, but are not actively annoying our users right now.

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -141,7 +141,7 @@ export default class Server {
         if (await this.context.isLWCTemplate(doc)) {
             this.auraDataProvider.activated = false; // provide completions for lwc components in an Aura template
             this.lwcDataProvider.activated = true;
-            if (params.context?.triggerCharacter === '{') {
+            if (this.shouldProvideBindingsInHTML(params)) {
```
randi274, line 144: Open to better naming here and for the `shouldCompleteJavascript` below, but I definitely didn't think the previous statement was clear on why we were matching on specific characters.

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
```diff
@@ -76,26 +81,44 @@ describe('handlers', () => {
...
+        it('should not return a list of completion items in a javascript file for open curly brace', async () => {
+            const params: CompletionParams = {
+                textDocument: { uri: jsUri },
+                position: {
+                    line: 0,
+                    character: 0,
+                },
+                context: {
+                    triggerCharacter: '{',
+                    triggerKind: 2,
```
gbockus-sf, line 112: total nit: do we have an enum for triggerKind?

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
```diff
@@ -10,6 +10,10 @@ const filename = path.resolve('../../test-workspaces/sfdx-workspace/force-app/ma
 const uri = URI.file(filename).toString();
 const document: TextDocument = TextDocument.create(uri, 'html', 0, fsExtra.readFileSync(filename).toString());
 
+const jsFilename = path.resolve('../../test-workspaces/sfdx-workspace/force-app/main/default/lwc/todo/todo.js');
+const jsUri = URI.file(jsFilename).toString();
+const jsDocument: TextDocument = TextDocument.create(uri, 'javascript', 0, fsExtra.readFileSync(jsFilename).toString());
```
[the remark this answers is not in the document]
randi274: Cool, good to know! Line 48 is the only place we need this addition, so I added it here for consistency

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
randi274: I actually think we should make an additional WI spike to figure out what how and why those are included. They don't seem particularly helpful, and if anything are also excessively noisy.

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
randi274: So this comes from the vscode-language-server-protocol, where it looks like CompletionTriggerKind is declared as a type. Not sure if we can/want to import that here.

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -150,17 +150,19 @@ export default class Server {
...
+                    isIncomplete: false,
+                    items: customTags,
                 };
-            });
-
-            return {
-                isIncomplete: false,
-                items: customTags,
-            };
+            }
```
gbockus-sf, line 165: so previously we always returned here, but we only return when it passes shouldCompleteJavascript.  Should we move the inner if into the above condition, or should we be returning with no results when no matching tag was found?

randi274: Ah, I see what you mean. I think an early empty return like what we have for the last else would be a good change here.

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
gbockus-sf: ah gotcha..yeah looks like we can just import and use that type. 
Screen Shot 2022-03-15 at 4 13 09 PM

packages/lwc-language-server/src/lwc-server.ts
randi274: Added!

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
randi274: Cool beans, just added!

packages/lwc-language-server/src/lwc-server.ts
gbockus-sf: things that make you go hummmmmmmmm.  How does undefined qualify as a CompletionList? 
weird. 

I did a little digging and think it's b/c `CompletionList` is a namespace and an interface (maybe).  sometimes I wonder if the teams at microsoft talk to each other :).  eh.  def better thanks for the update. 
Screen Shot 2022-03-15 at 4 53 13 PM

gbockus-sf (approved):
good work 👍

packages/lwc-language-server/src/__tests__/lwc-server.test.ts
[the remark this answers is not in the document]
jeffb-sfdc: I would have too.

jeffb-sfdc (approved):

PR #491, opened 2022-03-22
fix: fixed for:each error on hover

randi274:
### What does this PR do?
Addressed an issue where a for:each was not available, and was causing an undefined error.

To reproduce, hover over an item that doesn't have associated information with it, like the value of a for:each or a class name.

@W-10857157@, vscode #3929


packages/lwc-language-server/src/lwc-data-provider.ts
```diff
@@ -45,8 +45,8 @@ export class LWCDataProvider implements IHTMLDataProvider {
     }
     provideValues(): IValueData[] {
         const values: IValueData[] = [];
-        this.indexer.customData.forEach(t => {
-            t.classMembers.forEach(cm => {
+        this.indexer.customData?.forEach(t => {
+            t.classMembers?.forEach(cm => {
```
gbockus-sf, line 49: can we add a unit test for this condition? One to cover where customData isn't defined and one for when t.classMembers isn't defined?

randi274: I knew you were gonna ask for it 😆

jeffb-sfdc: I added a unit test for the case when t.classMembers is undefined.

For indexer.customData, indexer.customData() is a wrapper around `ComponentIndexer.tags`, and tags can never be null (in the body of the class, it's defined as `readonly tags: Map<string, Tag> = new Map();`).

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -116,7 +116,7 @@ it('transform throws exceptions on syntax errors', async () => {
         // verify err has the info we need
         const message = extractMessageFromBabelError(err.message);
         expect(message).toMatch('Unexpected token (4:17)');
-        expect(err.location).toEqual({ line: 4, column: 17 });
+        expect(err.location).toEqual({ line: 4, column: 17, index: 110 });
```
randi274, line 119: I'm not sure why this was failing, but the addition of the index is now expected here.

jeffb-sfdc (approved):

PR #496, opened 2022-05-02
fix: User input character should re-trigger autocomplete inside {} for HTML file

floralan:
### What does this PR do?
User type a character inside {} should re-trigger autocomplete in HTML file. 

@W-10930502@

### AC:
- need Ctrl + Space to wake up intellisense 
- once user type any characters within {}, autocomplete should be re-triggered.
- Ensure triggered *only* in HTML files (do not trigger in javascript)
- Ensure intellisense only triggered in HTML files that are in a Salesforce context
- Respect the VSCode ignore-intellisense setting ("editor.suggestOnTriggerCharacters": false, "editor.acceptSuggestionOnCommitCharacter": false)
- Clicking ctrl+space within curly brackets from HTML file should not trigger a different autocomplete list (W-10925630)

### DIFF
--- a/packages/lwc-language-server/src/lwc-server.ts
+++ b/packages/lwc-language-server/src/lwc-server.ts
@@ -140,7 +140,7 @@ export default class Server {
 
         if (await this.context.isLWCTemplate(doc)) {
             this.auraDataProvider.activated = false; // provide completions for lwc components in an Aura template
-            this.lwcDataProvider.activated = true;
+            this.lwcDataProvider.activated = true;       
             if (this.shouldProvideBindingsInHTML(params)) {
                 const docBasename = utils.getBasename(doc);
                 const customTags: CompletionItem[] = this.findBindItems(docBasename);
@@ -173,9 +173,20 @@ export default class Server {
         }
         return this.languageService.doComplete(doc, position, htmlDoc);
     }
-
+    
     private shouldProvideBindingsInHTML(params: CompletionParams): boolean {
-        return params.context?.triggerCharacter === '{';
+        const position = params.position;
+        const doc = this.documents.get(params.textDocument.uri);
+        const offset = doc.offsetAt(position);
+        const text = doc.getText();
+        let startIndex = offset - 1;
+        let char = text.charAt(startIndex);
+        const regPattern = /(\w|\$)/; // Valid variable names in JavaScript can contain letters, digits, underscore or $
+        while(char.match(regPattern)) {
+            startIndex -= 1;
+            char = text.charAt(startIndex);
+        }
+        return params.context?.triggerCharacter === '{' || char === '{'; 
     }
 
     private shouldCompleteJavascript(params: CompletionParams): boolean {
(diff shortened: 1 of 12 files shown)


packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -175,7 +175,18 @@ export default class Server {
     }
 
     private shouldProvideBindingsInHTML(params: CompletionParams): boolean {
-        return params.context?.triggerCharacter === '{';
+        const position = params.position;
```
randi274, line 178: @floralan can you move this logic into another function? I'd like for this to be a little easier to skim and understand. I'm thinking lines 178-188 could all be something different, since you're checking for if char ==='{'. Specifically it would be better to call out, what is char here, and we do we care about it?

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -175,7 +175,18 @@ export default class Server {
...
+        const doc = this.documents.get(params.textDocument.uri);
+        const offset = doc.offsetAt(position);
+        const text = doc.getText();
+        let startIndex = offset - 1;
+        let char = text.charAt(startIndex);
+        const regPattern = /(\w|\$)/; // Valid variable names in JavaScript can contain letters, digits, underscore or $
```
randi274, line 184: Is it possible to add additional tests to check for these patterns?

packages/lwc-language-server/src/lwc-server.ts
floralan: Yes, this makes sense to me. Just created a new function to it.

packages/lwc-language-server/src/lwc-server.ts
floralan: @randi274 Tests have been updated as well.

randi274 (commented):
I'm running through a few more manual tests, and will plan to wrap that up in the morning. Overall, looking great! 🎉

randi274 (approved):
Testing confirmed:
✅ Autocompletion still works in the HTML with { as the trigger character
✅ Nothing changed with the { in a JS, CSS, or Apex file
✅ Retries now work in the HTML w/ the ctrl + space
✅ Auto-complete within curly braces is now specific to properties in the JS

❓ - Clicking ctrl + space elsewhere within the HTML we still get that strange auto-complete. Are we ok with that? Do we want to open a new ticket, or is it ok as-is? Not a blocker, just something to consider.

floralan:
> Testing confirmed: ✅ Autocompletion still works in the HTML with { as the trigger character ✅ Nothing changed with the { in a JS, CSS, or Apex file ✅ Retries now work in the HTML w/ the ctrl + space ✅ Auto-complete within curly braces is now specific to properties in the JS
> 
> ❓ - Clicking ctrl + space elsewhere within the HTML we still get that strange auto-complete. Are we ok with that? Do we want to open a new ticket, or is it ok as-is? Not a blocker, just something to consider.

Re for ❓: Yes, this is the default autocomplete from VSCode, It will be overwritten if we have our own defined autocomplete functionalities for particular characters like `{}` in this case. I don't think we have the ability to disable it for now since we use vscode-html-languageservice, and I also think it's better to have it since it has no bad impacts to our current autocomplete functionalities. Ctrl + Space returns a list of all the **string** in the current HTML file, and users may need them somehow if they need reuse a string, which depends on their coding habits.

randi274:
Works for me @floralan!

PR #499, opened 2022-05-12
chore: update lwc

nolanlawson:
### What does this PR do?

Updates LWC dependencies from 1.7.14 to 2.13.3.

Note the breaking change, which is that `@lwc/engine` was renamed to `@lwc/engine-dom`. I've fixed the references to `@lwc/engine`.

This fixes several issues, e.g. that certain warnings or errors that were added after 1.7.14 do not surface in the VS Code extension. Now, they do. For example, duplicate `</template></template>` tags (https://github.com/salesforce/lwc/pull/2649) now show a warning (yellow squiggle):

Screen Shot 2022-05-12 at 3 10 33 PM

N/A

### DIFF
--- a/packages/lwc-language-server/src/template/__tests__/linter.test.ts
+++ b/packages/lwc-language-server/src/template/__tests__/linter.test.ts
@@ -2,18 +2,23 @@ import { TextDocument } from 'vscode-languageserver';
 import templateLinter from '../linter';
 
 it('returns a list of all the template compilation errors', () => {
-    const content = `<template><template if:true="invalidExpression">{Math.random()}</template><lighting-card></lighting-card></template>`;
+    const content = '<template><template if:true="invalidExpression"></template>{Math.random()}<lighting-card></lighting-card></template>';
     const document = TextDocument.create('test://test.html', 'html', 0, content);
 
     const diagnostics = templateLinter(document);
-    expect(diagnostics).toHaveLength(2);
+    expect(diagnostics).toHaveLength(3);
     expect(diagnostics[0].message).toMatch(/If directive should be an expression/);
     expect(diagnostics[0].range).toMatchObject({
         start: { character: 20 },
         end: { character: 47 },
     });
-    expect(diagnostics[1].message).toMatch(/<lighting- is not a valid namespace, sure you didn't mean "<lightning-"?/);
+    expect(diagnostics[1].message).toMatch(/Invalid expression/);
     expect(diagnostics[1].range).toMatchObject({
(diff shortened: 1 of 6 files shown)


packages/lwc-language-server/src/template/__tests__/linter.test.ts
```diff
@@ -2,7 +2,7 @@ import { TextDocument } from 'vscode-languageserver';
 import templateLinter from '../linter';
 
 it('returns a list of all the template compilation errors', () => {
-    const content = `<template><template if:true="invalidExpression">{Math.random()}</template><lighting-card></lighting-card></template>`;
+    const content = '<template><template if:true="invalidExpression"></template>{Math.random()}<lighting-card></lighting-card></template>';
```
nolanlawson, line 5: It looks like LWC bails out early now because the `if:true` part is invalid and doesn't check the contents of the `if:true` block. Moving the `{Math.random()}` outside of the `if:true` block fixes this.

randi274:
Wonderful, thanks for cleaning up our LWC dependencies @nolanlawson! I was actually about to work on a ticket to do the same, and the only part left for me is getting the LWC compiler up to date. I noticed you didn't add that here; was that out of scope for the work you needed to do, or did you run into any issues? I'm seeing a few now, but thought I'd check to see if you know of any specific reasons we might want to hold off on updating lwc/compiler.

randi274 (approved):

nolanlawson:
@randi274 I might have just missed that. Where is the compiler dependency defined?

nolanlawson:
Nevermind, I found it:

https://github.com/forcedotcom/lightning-language-server/blob/8f0ea20fee7ccf4a50a3ca4337dffd75d2bdd47e/packages/lwc-language-server/package.json#L26

Unfortunately it looks like this might be tricky, because you would be updating from `@lwc/compiler` v0.34.8 to v2.13.3, which dropped support for metadata (https://github.com/salesforce/lwc/pull/1016), which you are using in various places, e.g. here:

https://github.com/forcedotcom/lightning-language-server/blob/8f0ea20fee7ccf4a50a3ca4337dffd75d2bdd47e/packages/lwc-language-server/src/javascript/compiler.ts#L132

We might want to sync offline about how we can get `lightning-language-server` migrated.

randi274:
@nolanlawson that would be great actually - I'm close to having a PR ready for review that I think meets our needs, but would also love if you could review it to make sure I'm not missing anything that's clear from the LWC library.

Also going to close this specific PR as the work has now been merged in - thanks again!

Issue #493, opened 2021-09-22
Viewing LWC code produces errors in `LWC Language Server` tab


az-oolloow:
### Summary

_Short summary of what is going on or to provide context_.

### Steps To Reproduce:

1. Open a LWC js file (sometimes takes a couple of files)
2. Switch to LWC Language Server output tab

### Expected result

There are no errors if your component is composed correctly

### Actual result

```
[Error - 11:16:49 AM] Request textDocument/definition failed.
  Message: Request textDocument/definition failed with message: Cannot read property 'find' of undefined
  Code: -32603 
```

### Additional information

**Salesforce Extension Version in VS Code**: v52.13.0

**SFDX CLI Version**: sfdx-cli/7.116.2 darwin-x64 node-v14.17.5

**OS and version**: macOS Big Sur 11.6

klewis-sfdc:
Closed per merge of https://github.com/forcedotcom/salesforcedx-vscode/pull/4216

PR #505, opened 2022-06-02
fix: handle malformed tags quietly

randi274:
### What does this PR do?
Suppresses a noisy exception that occurs when trying to use the Hover or Go To Definition functionality on a component that we are not able to compile correctly due to the outdated LWC Compiler. One example of this error occurs when a component is using the NavigationMixin. This error is now logged quietly in the LWC output panel instead, so we can receive better reports from our users around the type of components that are not being handled correctly.

#493 

### Notes
To test, open the ebikes-LWC repo, or another project using the NavigationMixin. For the ebikes repo, open the productListItem, hover over the {product.MSRP__C} (or any other attribute) and press Cmd to try and click through to the definition. The error `Request textDocument/definition failed. Message: Request textDocument/definition failed with message: Cannot read properties of undefined (reading 'find')` will appear. Reload with the changes, and verify that there is no pronounced error grabbing focus. The error is logged in the LWC output tab instead:

Screen Shot 2022-06-02 at 4 15 25 PM

For an example of how the feature is supposed to work, open the accountMap and Cmd+Click to go to the definition of any attribute (like `markers`).

### DIFF
--- a/packages/lwc-language-server/src/tag.ts
+++ b/packages/lwc-language-server/src/tag.ts
@@ -117,7 +117,7 @@ export default class Tag implements ITagData {
     }
 
     classMember(name: string): ClassMember | null {
-        return this.classMembers.find(item => item.name === name) || null;
+        return this.classMembers?.find(item => item.name === name) || null;
     }
 
     classMemberLocation(name: string): Location | null {
--- a/packages/lwc-language-server/src/tag.ts
+++ b/packages/lwc-language-server/src/tag.ts
@@ -123,6 +123,7 @@ export default class Tag implements ITagData {
     classMemberLocation(name: string): Location | null {
         const classMember = this.classMember(name);
         if (!classMember) {
+            console.log(`Attribute "${name}" not found`);
             return null;
         }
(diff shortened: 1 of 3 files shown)


packages/lwc-language-server/src/tag.ts
```diff
@@ -117,12 +117,13 @@ export default class Tag implements ITagData {
...
 
     classMember(name: string): ClassMember | null {
-        return this.classMembers.find(item => item.name === name) || null;
+        return this.classMembers?.find(item => item.name === name) || null;
     }
 
     classMemberLocation(name: string): Location | null {
         const classMember = this.classMember(name);
         if (!classMember) {
+            console.log(`Attribute "${name}" not found`);
```
klewis-sfdc, line 126: @sbudhirajadoc this is the user-facing text that will be shown in the Output tab when the LWC Language Server option is selected.

klewis-sfdc:
✅ Looks good! The output on the LWC Language Server tab is more helpful than before and avoids the more abstract error that was being surfaced before.  Thanks @randi274 !

klewis-sfdc (approved):

PR #514, opened 2022-10-13
refactor: transition to github actions

CristiCanizales:
### What does this PR do?
Removes CircleCI configuration, documentation, and tasks in favor of Github Actions.

New Github Workflows:

- testCommitExceptMain.yml: Runs unit tests on every commit, except to main.
- onPushToMain.yml: On push to main, creates a github version, tag, and release.
- onRelease.yml: On successful creation of the release, publishes the library and the plugin to NPM.
- validatePR.yml: A PR can only be merged to main if all unit tests pass.

@W-11596519@


.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,48 @@
...
+  release:
+    runs-on: ubuntu-latest
+    container:
+      image: node:lts
+    needs:
+    - tests
+    steps:
+    - uses: actions/download-artifact@v2
+    - uses: actions/checkout@v3
+    - name: Configuring GitHub
```
randi274, line 19: You should be able to change this to invoke the github action above

randi274: Just make sure to pass in the ${{ secrets.GH_EMAIL }}

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,48 @@
...
+  tests:
+      uses: salesforcecli/github-workflows/.github/workflows/unitTest.yml@main
+  release:
+    runs-on: ubuntu-latest
+    container:
+      image: node:lts
+    needs:
+    - tests
+    steps:
+    - uses: actions/download-artifact@v2
```
randi274, line 17: This line we can remove, since we haven't used an upload-artifact action prior to this.

.github/workflows/slack_notify.yml
```diff
@@ -0,0 +1,27 @@
...
+      - Push to main
+      - release
+    types:
+      - completed
+
+jobs:
+  notify:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Announce result
```
randi274, line 14: Can you add another job if the status of the job is a failure? Something like: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow

.github/workflows/slack_notify.yml
```diff
@@ -0,0 +1,27 @@
...
+          payload: |
+            {
+              "text": "GitHub Actions Notification",
+              "event": "${{ github.event.workflow_run.name }}, run: ${{ github.event.workflow_run.html_url }}",
+              "repo": "${{ github.event.workflow_run.repository.name }}",
+              "result": "${{ github.event.workflow_run.conclusion }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.IDEE_RELEASE_ALERT_SLACK_WEBHOOK }}
+          SLACK_CHANNEL: pdt_releases
```
randi274, line 26: Is the channel necessary here?

.github/workflows/testCommitExceptMain.yml
```diff
@@ -0,0 +1,8 @@
+name: unit tests
+on:
+  push:
+    branches-ignore: [main]
```
randi274, line 4: I like this setup, where you're able to make sure tests pass again before merging to main in the onPushToMain workflow 👍

.vscode/tasks.json
```diff
@@ -1,12 +1,6 @@
...
   "version": "2.0.0",
   "tasks": [
-    {
-      "label": "Publish Language Servers",
-      "command": "./scripts/publish-workflow.sh",
-      "type": "shell",
-      "args": ["${input:circleCiId}"]
-    },
     {
       "label": "Create Port PR for Publishing",
```
randi274, line 5: I believe this Port PR for Publishing can also be removed

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,48 @@
...
+    - uses: actions/checkout@v3
+    - name: Configuring GitHub
+      run: |-
+        git config credential.helper 'cache --timeout=120'
+        git config user.email "$GH_EMAIL"
+        git config user.name "Release Bot"
+    - name: Bump package version
```
randi274, line 24: Instead of this step, can you add in the conventional changelog action?

https://github.com/forcedotcom/salesforcedx-templates/blob/777e382e78f457d631c3a55fe3956ee2763aac90/.github/workflows/onPushToMain.yml#L20

You'll want to use the whole step from templates, except for the fallback-version set that to the current version of the LSP (./lerna.json version). 

Then in your bump-lerna-version script, insert lines 26, 27, 28 (yarn, yarn bump-versions with the version from the changelog action (see templates bump-version script), and git add . )  

And remove the "Tag the release and push to repo step", since that will be covered with the changelog action.

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,22 @@
...
+  push:
+    tags:
+      - v*
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    container:
+      image: node:lts
+    steps:
+    - uses: actions/download-artifact@v2
```
randi274, line 13: This can be removed as well

.github/workflows/onPushToMain.yml
```diff
@@ -0,0 +1,48 @@
...
+      run: |-
+        yarn
+        yarn bump-versions 'minor'
+        git add .
+        export RELEASE_TAG="$(node -pe "require('./lerna.json').version")"
+        git commit -m "Updated version $RELEASE_TAG [ci skip]"
+    - name: Set .npmrc
```
randi274, line 31: Since this is in onRelease, Set .npmrc and the yarn publish-lsp and git push origin main steps can all be removed here

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,22 @@
...
+    - uses: actions/checkout@v2
+  npm:
+    runs-on: ubuntu-latest
+    container:
+      image: node:lts
+    steps:
```
randi274, line 19: The one thing I'm thinking might be missing here is the ssh keys... @gbockus-sf can you think of reason we would need those here for the publish to work?

randi274 (changes requested):
Wooo great work on this! Got a few changes now that I'm able to see the workflows end to end. Super close to getting this one migrated though 👏👏

.github/workflows/slack_notify.yml
CristiCanizales: It wasn't! Thank you :)

.github/workflows/slack_notify.yml
CristiCanizales: Discussed in slack :)
- This is going to be done in the second phase of slack notifications :) as well as other details

.github/workflows/onRelease.yml
gbockus-sf: Reaching back to remember how this works.  We need the ssh keys when we need to merge into protected branch like main.  If for this release process we don't have to do any commits we probably don't need the ssh keys.

.github/actions/setup-publish/action.yml
```diff
@@ -0,0 +1,24 @@
+name: setup-publish
+description: Build and package to prepare for publishing.
```
randi274, line 2: As discussed, the whole setup-publish directory can be removed

randi274 (approved):

.github/workflows/onRelease.yml
```diff
@@ -0,0 +1,21 @@
...
+    - uses: actions/checkout@v2
+  npm:
+    runs-on: ubuntu-latest
+    container:
+      image: node:lts
+    steps:
+      - name: Set .npmrc
+        run: echo "//registry.npmjs.org/:_authToken=${NPM_TOKEN}" > ~/.npmrc
```
randi274, line 20: @CristiCanizales I missed this line - can you change this to ${{ secrets.NPM_TOKEN }}, and check that that resolves ok in your forked repo?

randi274 (changes requested):

.github/workflows/onRelease.yml
randi274: actually... IDEE_NPM_TOKEN

.github/workflows/slack_notify.yml
```diff
@@ -0,0 +1,25 @@
...
+        with:
+          payload: |
+            {
+              "text": "GitHub Actions Notification",
+              "event": "${{ github.event.workflow_run.name }}, run: ${{ github.event.workflow_run.html_url }}",
+              "repo": "${{ github.event.workflow_run.repository.name }}",
+              "result": "${{ github.event.workflow_run.conclusion }}"
+            }
+        env:
+          SLACK_WEBHOOK_URL: ${{ secrets.IDEE_RELEASE_ALERT_SLACK_WEBHOOK }}
```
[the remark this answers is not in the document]
randi274: This is now available at the org level

randi274:
This is waiting on the commit-workflow, which should no longer be valid, so :merging:

PR #518, opened 2022-11-10
chore: update lwc versions

randi274:
### What does this PR do?
PR updates the lwc, engine, errors, and template-compiler versions to the latest in preparation for the next major release. Includes updates made to various warnings and errors and adds support for new directives added to the compiler. This change is needed for the 242 release.

### What issues does this PR fix or reference?
@W-11955917@, @W-11940409@

PR #521, opened 2022-11-10
chore: add a manual release as a backup when one isn't triggered by CI

randi274:
### What does this PR do?
Create a manual release process that will bump versions and tag a new release by default.

Allow for shipping #518


gbockus-sf (approved):
Looks good. Thanks for checking the camelCase conversion.

gbockus-sf:
We slacked about it, but also just putting it here for sharing sake.  The unit test check that is failing if mostly likely due to the lts/* version of node switching from v16-> v18 https://github.com/nodejs/Release

PR #522, opened 2022-11-11
feat: node to LTS to support github actions release

randi274:
### What does this PR do?
Bumps node to LTS-1 (v16.0.0) as we were having issues publishing with a greatly outdated node version.

Continuing improvements to #521.

BREAKING CHANGE: Major node version change requires devs to update to 16.0.*


gbockus-sf (approved):
🚢

.github/workflows/startManualRelease.yml
```diff
@@ -0,0 +1,37 @@
...
+        uses: TriPSs/conventional-changelog-action@d360fad3a42feca6462f72c97c165d60a02d4bf2
+        # Get the version bump from Changelog Action + Create Tag
+        with:
+          git-user-name: Release Bot
+          git-user-email: ${{ secrets.IDEE_GH_EMAIL }}
+          github-token: ${{ secrets.IDEE_GH_TOKEN }}
+          skip-version-file: true
+          skip-on-empty: false
+          output-file: false
+          pre-commit: ./scripts/bump-lerna-version.js
```
gbockus-sf, line 28: hummmmm insteresting.  So the conventional-changelog is also going to try and bump the version in package.json.  Does skip-version-file take care of that for us I guess?  Does bump-lerna-version also do a commit to commit all the packge.*json updates?

I'm wondering if we just need to do the version bump in learn as a separate step that happens before (or maybe even instead of) the change log thing.

randi274: So the reason I like to have it here is so the math is the same for the version + the git tag. Conventional changelog decides what the version should be and passes it to the bump-lerna-version script for us to read and just set, rather than having to decide that separately. We don't seem to have any option to set the version into conventional changelog... so this is the way.

gbockus-sf: gotcha, makes sense.  I was just looking at the error and can't understand why it's beefing about the export in that preCommit script.

PR #528, opened 2022-11-28
lightning/analyticsWaveApi typings updates

smithgp:
### What does this PR do?

Add typings for `lightning/analyticsWaveApi` adapters that have been added since the 238 release. These are all already public and doc'ed. Most of the content here is taken from the types generated by the adapter build and the public docs.

@W-10758280@

cc: @mblumreich, @karihotchkiss, @reidaelliott, @rdubovetskyi, @lindsey-burnett


randi274:
Done and merged with #530  - thanks @smithgp!

PR #538, opened 2023-01-11
chore: lwc v2.35.2 @W-12246411

ekashida:
Update lwc v2.35.2


klewis-sfdc:
Hi @ekashida,  Thanks for submitting these dependency updates.  It looks like the updates to the lockfile have the modules being installed from Nexus, and the GitHub Actions fail as they don't have access.  Are you able to update the lockfile so that these dependencies are installed from the NPM registry instead?

PR #549, opened 2023-02-28
chore: match startManualRelease and onPushToMain release jobs

randi274:
### What does this PR do?
The onPushToMain job, which creates a changelog and generates a github release, started failing recently, but the startManualRelease, which has almost the same set of steps, was still [working](https://github.com/forcedotcom/lightning-language-server/actions/runs/4295325354). The two jobs are now much more closely aligned, and the onPushToMain should now be passing as well. 

@W-12592699@
Closes forcedotcom/easywriter#138

### DIFF
--- a/.github/workflows/onPushToMain.yml
+++ b/.github/workflows/onPushToMain.yml
@@ -9,36 +9,32 @@ jobs:
     uses: salesforcecli/github-workflows/.github/workflows/unitTest.yml@main
   release:
     runs-on: ubuntu-latest
-    container:
-      image: node:lts
     needs:
     - tests
     steps:
-    - uses: actions/checkout@v3
-      with:
-        token: ${{ secrets.IDEE_GH_TOKEN }}
-    - uses: ./.github/actions/gitConfig
-      with:
-        email: ${{ secrets.IDEE_GH_EMAIL }}
-    - run: yarn
-    - name: Conventional Changelog Action
-      id: changelog
-      uses: TriPSs/conventional-changelog-action@ed98b658f3015c5afb96a13caf9bb81f69df49e6
-      with:
-        git-user-name: Release Bot
-        git-user-email: ${{ secrets.IDEE_GH_EMAIL }}
-        github-token: ${{ secrets.IDEE_GH_TOKEN }}
-        skip-version-file: true
-        output-file: false
-        tag-prefix: "v"
-        pre-commit: ./scripts/bump-lerna-version.js
-        fallback-version: 3.12.0
-    - name: Create Github Release
-      uses: actions/create-release@v1
-      if: ${{ steps.changelog.outputs.skipped == 'false' }}
-      env:
-        GITHUB_TOKEN: ${{ secrets.IDEE_GH_TOKEN }}
-      with:
-        tag_name: ${{ steps.changelog.outputs.tag }}
-        release_name: ${{ steps.changelog.outputs.tag }}
-        body: ${{ steps.changelog.outputs.clean_changelog }}
+      - uses: actions/setup-node@v3
+        with:
+          node-version: '16.0.x'
+      - uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.IDEE_GH_TOKEN }}
+      - run: yarn
+      - name: Conventional Changelog Action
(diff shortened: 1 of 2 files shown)


.github/workflows/onPushToMain.yml
```diff
@@ -9,36 +9,32 @@ jobs:
...
     runs-on: ubuntu-latest
-    container:
-      image: node:lts
     needs:
     - tests
     steps:
-    - uses: actions/checkout@v3
-      with:
-        token: ${{ secrets.IDEE_GH_TOKEN }}
-    - uses: ./.github/actions/gitConfig
```
randi274, line 20: The other job doesn't use this, so I removed the call to the gitConfig as well. I don't think anything else is using it, so I could remove this section too, or keep it in just in case.

.github/workflows/onPushToMain.yml
```diff
@@ -9,36 +9,32 @@ jobs:
     uses: salesforcecli/github-workflows/.github/workflows/unitTest.yml@main
   release:
     runs-on: ubuntu-latest
-    container:
-      image: node:lts
```
randi274, line 13: Based on some light [research](https://github.com/actions/checkout/issues/363#issuecomment-1073731046), I think this was our likely culprit.

.github/workflows/onPushToMain.yml
```diff
@@ -9,36 +9,32 @@ jobs:
...
-    - name: Create Github Release
-      uses: actions/create-release@v1
-      if: ${{ steps.changelog.outputs.skipped == 'false' }}
-      env:
-        GITHUB_TOKEN: ${{ secrets.IDEE_GH_TOKEN }}
-      with:
-        tag_name: ${{ steps.changelog.outputs.tag }}
-        release_name: ${{ steps.changelog.outputs.tag }}
-        body: ${{ steps.changelog.outputs.clean_changelog }}
+      - uses: actions/setup-node@v3
```
randi274, line 15: The other job also includes the setup-node, so I added this.

.github/workflows/onPushToMain.yml
```diff
@@ -9,36 +9,32 @@ jobs:
...
+        with:
+          node-version: '16.0.x'
```
gbockus-sf, line 17: `node-version-file: '.nvmrc'` FTW?

.github/workflows/onPushToMain.yml
```diff
@@ -9,36 +9,32 @@ jobs:
...
-    - name: Conventional Changelog Action
-      id: changelog
-      uses: TriPSs/conventional-changelog-action@ed98b658f3015c5afb96a13caf9bb81f69df49e6
-      with:
-        git-user-name: Release Bot
-        git-user-email: ${{ secrets.IDEE_GH_EMAIL }}
-        github-token: ${{ secrets.IDEE_GH_TOKEN }}
-        skip-version-file: true
-        output-file: false
-        tag-prefix: "v"
```
randi274, line 33: From this section, I removed tag-prefix and fallback version (which was necessary for the first run, but isn't any longer).

.github/workflows/startManualRelease.yml
```diff
@@ -27,7 +27,6 @@ jobs:
           skip-on-empty: false
           output-file: false
           pre-commit: ./scripts/bump-lerna-version.js
-          tag-prefix: ''
```
randi274, line 30: This is being ignored in favor of the default value ("v"), so I removed it as well.

.github/workflows/onPushToMain.yml
randi274: ohhhh yes, lemme do that for both

.github/workflows/onPushToMain.yml
gbockus-sf: good hunting

.github/workflows/onPushToMain.yml
randi274: done!

.github/workflows/onPushToMain.yml
gbockus-sf: better to be gone if it can be 👍

PR #552, opened 2023-03-01
ci: update manual release and push to main to allow for nvmrc usage

randi274:
### What does this PR do?
Swaps the order of checkout and node setup in 2 github actions so we can use the nvmrc of the checked out code to determine the node version used.

Follow-up fix to #549


gbockus-sf (approved):
haaaproved

PR #558, opened 2023-04-10
fix(core): update core config file and handle read errors

### LINKED ISSUE #497: LWC Language Server errors on load when config.blt is not present
### Summary

Salesforce VS Code Extensions fail to load due to the following error when a `config.blt` file is not present on the local machine. I just set up a new laptop with `corecli` and it looks like `config.blt` is no longer a required file for a core workspace.

We should accommodate this scenario by parsing an alternative config file when `config.blt` is not present.

```
[Error: ENOENT: no such file or directory, open '/Users/<user>/blt/config.blt'] {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: '/Users/<user>/blt/config.blt'
}
[Error - 3:57:12 PM] Connection to server got closed. Server will not be restarted.
```

I also see the following error popup in VS Code on load:

```
 The LWC Language Server server crashed 5 times in the last 3 minutes. The server will not be restarted.
```

### Steps To Reproduce:

1.  Set up core following the latest internal doc (do not use any BLT commands that would init a config.blt file).
2. Install the VS Code Extension Pack from the public marketplace.
3. Restart VS Code.

### Expected result

VS Code loads without error.

### Actual result

The errors above are present.

### Additional information

I see the parsing of `config.blt` done here with no error handling or fallback. We should find an alternative file to read if this line fails:

https://github.com/forcedotcom/lightning-language-server/blob/7cd0d5736b546fb4cab2963a5fb93ea284b8a0b6/packages/lightning-lsp-common/src/context.ts#L485

**VS Code Version**: 1.67.0 (Universal)

**SFDX CLI Version**: 

**OS and version**: macOS Monterey 12.3.1

mpsenn:
The core config file containing environment variables is now build/dev.properties. Use that file instead of config.blt. Additionally handle any errors in parsing and reading by logging to console and not attempting to write to the file.

Fixes [@W-13007946@](https://gus.lightning.force.com/lightning/r/ADM_Work__c/a07EE00001OplXFYAZ/view). Fixes #497.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@
    let you know if there is already an effort in progress.
 1. Fork this repository.
 1. The [README](README.md) has details on how to set up your environment.
-1. Create a _topic_ branch in your fork based on the correct branch (usually the **develop** branch, see [Branches section](#branches) below). Note, this step is recommended but technically not required if contributing using a fork.
+1. Create a _topic_ branch in your fork based on the correct branch (usually the **main** branch, see [Branches section](#branches) below). Note, this step is recommended but technically not required if contributing using a fork.
 1. Edit the code in your fork.
 1. Sign CLA (see [CLA](#cla) below)
 1. Send us a pull request when you are done. We'll review your code, suggest any
@@ -26,10 +26,10 @@ Agreement. You can do so by going to https://cla.salesforce.com/sign-cla.
 
 ## Branches
 
-- We work in `develop`.
+- We work in `main`.
 - Our released (aka. _production_) branch is `main`.
 - Our work happens in _topic_ branches (feature and/or bug-fix).
-  - feature as well as bug-fix branches are based on `develop`
+  - feature as well as bug-fix branches are based on `main`
   - branches _should_ be kept up-to-date using `rebase`
   - see below for further merge instructions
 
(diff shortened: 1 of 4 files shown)


mpsenn:
Fixes #497.

mpsenn:
Note: for git checkout of core, this generates settings with empty p4.port values, etc. This is fine, I think the eslint executable is what really matters.

jasonschroeder-sfdc (approved):

randi274 (changes requested):
QA looks good! For our SFDX users, we see the language server boots up as expected, and the usual functionality like hover, go-to-definition and auto-complete work great.

I am seeing some failing tests in our internal PR, and they're reproducible locally. Likely just a stubbing issue that needs to be updated given the new path. Let me know when you're able to take a look and I'll get the updated changes merged into my branch. Thanks!

https://github.com/forcedotcom/lightning-language-server/actions/runs/4680070576/jobs/8290938179

mpsenn:
@randi274 I've updated the change, after taking a closer look at a colleague's setup, it looks like env. variables are used for the config now instead of a properties file. So I've updated the implementation, and modified the test workspace to match.

randi274:
Thanks again @mpsenn! We'll get this updated in vs code ASAP 🎉

PR #564, opened 2023-04-12
fix: update core config file and handle read errors

### LINKED ISSUE #497: LWC Language Server errors on load when config.blt is not present

randi274:
### What does this PR do?
The core config file containing environment variables is now build/dev.properties. Use that file instead of config.blt. Additionally handle any errors in parsing and reading by logging to console and not attempting to write to the file.
Also updates documentation for outdated `develop` branch references.

@W-13007946@. Fixes #497. Closes #558.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,7 +7,7 @@
    let you know if there is already an effort in progress.
 1. Fork this repository.
 1. The [README](README.md) has details on how to set up your environment.
-1. Create a _topic_ branch in your fork based on the correct branch (usually the **develop** branch, see [Branches section](#branches) below). Note, this step is recommended but technically not required if contributing using a fork.
+1. Create a _topic_ branch in your fork based on the correct branch (usually the **main** branch, see [Branches section](#branches) below). Note, this step is recommended but technically not required if contributing using a fork.
 1. Edit the code in your fork.
 1. Sign CLA (see [CLA](#cla) below)
 1. Send us a pull request when you are done. We'll review your code, suggest any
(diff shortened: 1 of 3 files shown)


CristiCanizales (approved):
🚢

PR #566, opened 2023-04-27
feat(common): updates to analyticsWaveApi typings

smithgp:
### What does this PR do?
1. Adds `ids` to `getDatasets`
2. Adds `validateWaveTemplate`

These are for changes made the adapters in 58.0.

@W-13089840@

cc:  @mblumreich

### DIFF
--- a/packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
+++ b/packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
@@ -2061,6 +2061,65 @@ declare module 'lightning/analyticsWaveApi' {
     }
 
     /**
+     * An Analytics template asset reference.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_responses_asset_reference.htm
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface TemplateAssetReferenceRepresentation {
+        /** The 18 character ID of the asset. */
+        id?: string;
+        /** The asset label. */
+        label?: string;
+        /** The asset developer name. */
+        name?: string;
+        /** The namespace that qualifies the asset name */
+        namespace?: string;
+        /** The asset URL. */
+        url?: string;
+    }
+
+    /**
+     * Representation for individual validation task
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_responses_template_readiness_item.htm
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface TemplateReadinessItemRepresentation {
+        /** The icon/image associated with the validation task. */
+        image: TemplateAssetReferenceRepresentation | null;
+        /** The task specific label. */
+        label: string | null;
+        /** The task specific message. */
+        message: string | null;
+        /** The status for the readiness task */
+        readinessStatus: string | null;
+        /** The collection of tags describing the purpose of the validation. */
+        tags: Array<string>;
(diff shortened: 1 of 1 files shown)


packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -2062,6 +2060,65 @@ declare module 'lightning/analyticsWaveApi' {
...
+     */
+    export interface TemplateReadinessItemRepresentation {
+        /** The icon/image associated with the validation task. */
+        image: TemplateAssetReferenceRepresentation | null;
+        /** The task specific label. */
+        label: string | null;
+        /** The task specific message. */
+        message: string | null;
+        /** The status for the readiness task */
+        readinessStatus: string | null;
```
aarondailsf2023, line 2100: Since this is like an enum, could it use string literals in the type, like `readinessStatus: "failed" | "success" | null` (just making up those values, not sure if they are the right ones)?

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -2062,6 +2060,65 @@ declare module 'lightning/analyticsWaveApi' {
...
+    /**
+     * Representation for a single Wave Template verification.
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_responses_template_validate.htm
+     *
+     * Keys:
+     *    id (string): id
+     */
+    export interface TemplateValidateRepresentation {
+        /** The ID or fully qualified API name of this template. */
```
aarondailsf2023, line 2116: Nit: is this comment right? It can be id or name?

aarondailsf2023 (approved):

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
smithgp: These types come from the generated adapter code, and are basically copied here. And, we need to match what the adapter code says, since any usages of this in js will call through the adapter. In the adapter code, it (probably) intentionally doesn't pull union types for underlying enums, but just does `string`, so we do that here, too. We sometimes put a note in the jsdoc about the possible values, but generally just rely on the docs, since those get updated per release.

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
```diff
@@ -2062,6 +2060,65 @@ declare module 'lightning/analyticsWaveApi' {
...
+     * Representation for individual validation task
+     *
+     * https://developer.salesforce.com/docs/atlas.en-us.bi_dev_guide_rest.meta/bi_dev_guide_rest/bi_responses_template_readiness_item.htm
+     *
+     * Keys:
+     *    (none)
+     */
+    export interface TemplateReadinessItemRepresentation {
+        /** The icon/image associated with the validation task. */
+        image: TemplateAssetReferenceRepresentation | null;
```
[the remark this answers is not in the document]
smithgp: This is intentional, and again from the generated adapter typings. The adapter runtime does differentiate between optional and required-but-can-be-null, and it does match up with what the response from the server code will have. In this case, `image` will also be in the response json but can be null, which is why it's this way.

packages/lightning-lsp-common/src/resources/sfdx/typings/lds.d.ts
smithgp: Yep. A number of our rest apis (and therefore adapters) support that in the urls and then the resulting output json.

RitamAgrawal (approved):

PR #567, opened 2023-05-12
perf: improve time taken to init

mwaddoupffdc:
### What does this PR do?
Performance enhancement to speed up initialisation, particularly when working with large code bases. Tested locally, this has reduced the startup time for the extension from > 2 hours to around 2 minutes.

https://github.com/forcedotcom/salesforcedx-vscode/issues/4829


randi274 (approved):
Looks good and tests well for me! We'll get this merged in with pr/567, and updated in VS Code soon as well. Thanks for the contribution!

PR #570, opened 2023-06-23
feat: transition to @lwc/metadata for LWC JS introspection

divmain:
### What does this PR do?

This PR removes the `lwc-language-server` dependency on an ancient version of the LWC compiler. Now, `@lwc/metadata` is leveraged to provide intelligence on LWC JavaScript in the IDE.

Initially, the plan was to entirely swap out the data structures underlying a lot of the logic in the language server. Instead of relying on the very old `Metadata` type, all logic would be updated to instead rely on new metadata types exposed by `@lwc/metadata`. However, the changes necessary to make this clean swap proved to be extremely invasive. Continuing down that path would have required a de facto rewrite, with all the inevitable subtle incompatibilities and bugs that come with such an endeavor.

As an alternative, we chose to provide a mapping of `@lwc/metadata` to the old metadata data structure, including all of its idiosyncrasies. This approach comes with some benefits:

- the pre-existing behavior is reproduced ~exactly (or as close as we could come to it)
- further iterations and refactors can leverage new LWC metadata without disrupting existing functionality
- existing tests could be leveraged, adding confidence that this PR includes only non-breaking changes
- it should allow `lwc-language-server` to stay up-to-date with future releases of LWC, with the possibility that small changes to the mapping might be rarely required

As a follow-up I'll work to configure our CI to run your test suite whenever we update `@lwc/metadata`. If successful, the LWC team will be alerted whenever we make breaking changes to your code. In such an event, we would either 1) avoid the break or 2) proactively inform you of the change.

@W-13045347@


package.json
```diff
@@ -30,7 +30,7 @@
     "patch-package": "^6.0.5",
     "rimraf": "^3.0.1",
     "shelljs": "^0.8.3",
-    "typescript": "^3.2.4"
+    "typescript": "^5.0.4"
```
divmain, line 33: Upgrading TypeScript was necessary in order to consume types exposed by `@lwc/metadata`.

packages/aura-language-server/package.json
```diff
@@ -41,26 +41,27 @@
...
     "lint-staged": "^10.2.11",
     "mock-fs": "^4.11.0",
     "npm-run": "^5.0.1",
     "prettier": "^2.0.5",
     "shelljs": "^0.8.4",
-    "ts-jest": "^26.2.0",
+    "ts-jest": "^29.1.0",
     "tsc-transpile-only": "0.0.3",
-    "typescript": "3.7.5"
+    "typescript": "5.0.4"
```
divmain, line 65: Dependency bumps here and `lightning-lsp-common` were necessary due to the top-level TypeScript upgrade to v5.

packages/lwc-language-server/package-lock.json
divmain, line 1: This monorepo has a top-level `yarn-lock` that conflicted with this package-local `package-lock.json`. Since the sibling packages did not have their own `package-lock.json`, I discerned that its existence here was probably unintentional.

packages/lwc-language-server/package.json
```diff
@@ -48,22 +50,25 @@
     "xml2js": "^0.4.23"
   },
   "devDependencies": {
+    "@jest/reporters": "^29.5.0",
+    "@komaci/types": "^246.0.2",
+    "@lwc/old-compiler": "npm:@lwc/compiler@0.34.8",
```
divmain, line 55: `lightning-language-server` relied upon this ancient version of the LWC compiler. It has been made a dev dependency here, allowing for the comparison of mapped metadata to original metadata.

packages/lwc-language-server/src/decorators/index.ts
```diff
@@ -46,6 +46,7 @@ export interface WireDecoratorTarget {
     params: { [name: string]: string };
     static: any;
     type: DecoratorTargetType;
+    adapter?: unknown;
```
divmain, line 49: This field is not used within `lwc-language-server`. However, it _is_ present in the old metadata.

We want to ensure that mapped metadata and old metadata are identical. To that end, we've added the field here to allow its inclusion in the mapped metadata.

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -15,86 +14,18 @@ import {
...
-
-    const options: CompilerOptions = {
-        name: 'foo',
-        namespace: 'x',
-        files: {},
-    };
-    const transformerResult = await transform(codeOk, 'foo.js', options);
-    const code = transformerResult.code;
-    expect(pretify(code)).toBe(pretify(expected));
-});
```
divmain, line 56: A number of these tests made assertions directly against the ancient LWC compiler. That's not particularly useful. It also cannot be left in place, since metadata is no longer exposed via the LWC compiler but through `@lwc/metadata` instead.

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -103,74 +34,35 @@ export default class Foo extends LightningElement {
...
-        fail('expects exception');
-    } catch (err) {
-        // verify err has the info we need
-        const message = extractMessageFromBabelError(err.message);
-        expect(message).toMatch('Unexpected token (4:17)');
-        expect(err.location).toEqual({ line: 4, column: 17 });
-    }
+it('can get metadata from a simple component', async () => {
+    await compileSource(codeOk, 'foo.js');
 });
```
divmain, line 39: We've added a test case for a bare-bones LWC component here. This does little more than ensure that the plumbing works and that errors are not thrown.

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -103,74 +34,35 @@ export default class Foo extends LightningElement {
...
-        const location = extractLocationFromBabelError(err.message);
-        expect(location).toEqual({ line: 5, column: 4 });
-    }
+it('displays an error for a component with syntax error', async () => {
+    const result = await compileSource(codeSyntaxError, 'foo.js');
+    expect(result.metadata).toBeUndefined();
+    expect(result.diagnostics).toHaveLength(1);
+    const [diagnostic] = result.diagnostics;
+    expect(diagnostic.message).toMatch('Unexpected token (4:17)');
 });
```
divmain, line 47: This is a new test that checks for diagnostics on a broken component, and corresponds most closely with the removed test labeled `it('transform throws exceptions on syntax errors')`.

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -215,68 +107,96 @@ it('transform returns javascript metadata', async () => {
     const filepath = path.join('src', 'javascript', '__tests__', 'fixtures', 'metadata.js');
```
divmain, line 107: We significantly augmented the fixture used in this test. Every variety of decorator that we could think of was added. That is reflected in the changes below.

packages/lwc-language-server/src/javascript/__tests__/compiler.test.ts
```diff
@@ -211,72 +103,98 @@ it('linter returns empty diagnostics on correct file', async () => {
     expect(diagnostics).toEqual([]);
 });
 
-it('transform returns javascript metadata', async () => {
+it('mapLwcMetadataToInternal returns expected javascript metadata', async () => {
```
divmain, line 106: We significantly augmented the fixture used in this test. Every variety of decorator that we could think of was added, which is reflected in the changes below.

packages/lwc-language-server/src/__tests__/component-indexer.test.ts
```diff
@@ -12,9 +12,8 @@ const componentIndexer: ComponentIndexer = new ComponentIndexer({
     workspaceRoot,
 });
 
-beforeEach(async done => {
+beforeEach(async () => {
     await componentIndexer.init();
-    done();
```
divmain, line 17: This started failing with new version of Jest. `beforeEach` expects a function that either returns void (but calls `done`) or returns a `Promise`.

packages/lwc-language-server/src/__tests__/typing-indexer.test.ts
```diff
@@ -6,6 +6,16 @@ const typingIndexer: TypingIndexer = new TypingIndexer({
...
+// This is required in order to spyOn fsExtra functions in newer versions of
+// jest and ts-jest. Solution adapted from Jest docs here:
+// https://jestjs.io/docs/jest-object
+jest.mock('fs-extra', () => {
+  return {
+    __esModule: true,
+    ...jest.requireActual('fs-extra'),
+  };
+});
+
```
divmain, line 18: The test `it('should not create a customlabels typing file when a project has no custom labels')` was broken without this fix. It seemed as though TypeScript changed the way it compiled ES modules, using `Object.defineProperty` for each export, which interferes with Jest's mocking. Fortunately, the fix is relatively straightforward.

packages/lwc-language-server/src/__tests__/typing-indexer.test.ts
```diff
@@ -66,6 +76,11 @@ describe('TypingIndexer', () => {
 <CustomLabels xmlns="http://soap.sforce.com/2006/04/metadata"/>
 `;
             jest.spyOn(fsExtra, 'readFileSync').mockReturnValue(Buffer.from(xmlDocument));
+            jest.spyOn(typingIndexer, 'customLabelFiles', 'get')
+                .mockReturnValue([
+                    '/foo/bar/test-workspaces/sfdx-workspace/utils/meta/labels/CustomLabels.labels-meta.xml'
+                ]);
```
divmain, line 82: Under the hood, multiple `readFileSync` invocations occurred. First in `sfdxConfig` (indirectly invoked by `customLabelFiles`) and then again in `saveCustomLabelTypings`. Because of this, I'm surprised the test actually passed before.

Making this change seems to preserve the intent of the test, and provides the unit-level isolation such that only a single `readFileSync` invocation occurs.

packages/lwc-language-server/src/javascript/__tests__/type-mapping.test.ts
```diff
@@ -0,0 +1,38 @@
+import * as path from 'path';
+import { collectBundleMetadata, BundleConfig, ScriptFile } from '@lwc/metadata';
+import { transform } from '@lwc/old-compiler';
+import { CompilerOptions as OldCompilerOptions } from '@lwc/old-compiler/dist/types/compiler/options';
+import { mapLwcMetadataToInternal } from '../type-mapping';
+import { Metadata } from '../../decorators';
+import * as fs from 'fs-extra';
+
+it('can map new metadata to old metadata', async () => {
```
divmain, line 9: This test ensures that `mapLwcMetadataToInternal(collectBundleMetadata(filename))` produces the exact same output as `transform(...). metadata`, bug-for-bug.

As a first step, this seemed like the best way to ensure introduction of `@lwc/metadata` would not result in bugs in the language server. Over time, if the language server is refactored to leverage `@lwc/metadata`'s output more directly, this test can be relaxed or removed altogether.

packages/lwc-language-server/src/javascript/compiler.ts
```diff
@@ -121,20 +124,43 @@ function toDiagnostic(err: any): Diagnostic {
...
-            name,
-            namespace: 'x',
-            files: {},
-        };
-        const transformerResult = await transform(source, fileName, options);
-        const metadata = transformerResult.metadata as Metadata;
-        patchComments(metadata);
-        return { metadata, diagnostics: [] };
+        transformSync(source, fileName, transformOptions);
     } catch (err) {
```
divmain, line 135: Some compilation errors will only appear upon an attempted compilation. These are generally restricted to non-syntax errors, such as the fact that boolean public properties cannot be true by default (otherwise, how would the parent component disable the flag?).

So, we transform first using the up-to-date compiler to ensure no compilation issues exist. If it succeeds, we collect metadata about the component code below.

packages/lwc-language-server/src/javascript/type-mapping.ts
```diff
@@ -0,0 +1,560 @@
+import {
```
divmain, line 1: This module transforms new metadata (from `@lwc/metadata`) into the data-structure produced by very old versions of the LWC compiler.

packages/lwc-language-server/src/javascript/type-mapping.ts
```diff
@@ -0,0 +1,560 @@
...
+
+    return {
+        name: prop.name,
+        type: 'property',
+        static: staticObj,
+        params,
+        // The old LWC metadata included an 'adapter' property, which
+        // is not used by lightning-language-server and is not captured in
+        // the WireDecoratorTarget type. But we've captured it here so that
+        // the output exactly matches that of the old LWC compiler's metadata.
```
divmain, line 395: To add to this comment, getting the `adapter` value is extremely inexpensive (as seen in the lines above). Because of that, it seemed worthwhile to include here with a corresponding change in the `WireDecoratorTarget` type. This allows for exact duplication of old metadata, making it easier to assert equivalency.

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -104,7 +104,7 @@ export default class Server {
             useDefaultDataProvider: false,
         });
 
-        this.context.configureProject();
+        await this.context.configureProject();
```
divmain, line 107: Parts of the test suite were failing because this wasn't awaited. I believe this is the only functional change in the PR, and doesn't appear to be problematic.

CristiCanizales (approved):
This looks good to me 🙂 And I know all the hard work to get to this 🥳 
I also tested the usual features were working as expected, like:
- go to definition (html) ✅ 
- go to definition (javascript) ✅ 
- autocompletion ✅ 
- info on hover ✅ 

So I would say 🚢

PR #574, opened 2023-07-06
fix: Update .nvmrc

CristiCanizales:
### What does this PR do?
bumps node engine to 16.15.1

@W-13512133@


.nvmrc
```diff
@@ -1 +1 @@
-v16.0.0
\ No newline at end of file
+v16.1.0
```
RitamAgrawal, line 1: Have you tried upgrading to v16.15.1?  This will keep the node version same across extensions and language server.

RitamAgrawal (approved):

PR #578, opened 2023-09-20
fix: language server would crash if main class could not be identified

divmain:
### What does this PR do?

In cases where the main class of a JavaScript file could not be identified, an error was thrown during analysis of the associated metadata. This would cause the language server to crash in VSCode. 👎 

This PR fixes the issue, providing metadata is there is an exported class or if only a single class is defined in the `.js` file. If multiple classes are defined and none of them are default exports, the language server will now return no data. No code hints will be provided, but the language server will not crash.

Fixes forcedotcom/salesforcedx-vscode#4994.


CristiCanizales:
This looks good to me, thank you for taking care of this!! :) We just need to get the tests passing and then we can :shipit:

PR #579, opened 2023-09-27
Cristi/unit tests

CristiCanizales:
### What does this PR do?
- Creates new workflow for unit tests that uses node version lts/-1 and lts/* and not latest and adds macos to the matrix

@W-14195245@


gbockus-sf (approved):
Full passing run [here](https://github.com/forcedotcom/lightning-language-server/actions/runs/6328447710)

Will update the GHA setting to address the workflow name change

PR #580, opened 2023-10-11
Make lwc & aura language server available to be run from an executable file

WuggyX2:
### What does this PR do?
This PR creates executable files and defines the bin property in package.json for both lwc language server and aura language server.

#581 

This allows the language servers to be run using a terminal command, which in return allows editors like neovim to integrate with language servers using `stdio`. [Neovim LSP DOC](https://neovim.io/doc/user/lsp.html#vim.lsp.start())

### DIFF
--- a/packages/aura-language-server/bin/aura-language-server.js
+++ b/packages/aura-language-server/bin/aura-language-server.js
@@ -1,0 +1,2 @@
+#!/usr/bin/env node
+require('../lib/server.js');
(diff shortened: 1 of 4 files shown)


gbockus-sf:
@WuggyX2 Have you attempted to take care of the salesforce-cla step above?

WuggyX2:
@gbockus-sf I tried, but it gave me an error, maybe because I used a different email than what is linked to this account. When I try to sign it again, the page say i have already signed it.

kuva

gbockus-sf:
@WuggyX2 Thanks for the follow up.  The check is passing now so you're good to go.  I'm just running this my our PM for awareness and it should be good to go I think.

gbockus-sf (approved):
Thanks @WuggyX2 merging!

PR #583, opened 2023-10-30
fix: update metadata mapping to handle no getter

gbockus-sf:
### What does this PR do?
Update the javascript type mapping to handle the case where a class has a property with only a setter.

@W-14391061@
https://github.com/forcedotcom/salesforcedx-vscode/issues/5069

### DIFF
--- a/packages/lwc-language-server/src/javascript/type-mapping.ts
+++ b/packages/lwc-language-server/src/javascript/type-mapping.ts
@@ -217,17 +217,25 @@ function getMemberProperty(propertyObj: ClassProperty): InternalClassMember | nu
         propertyObj.dataProperty,
     );
 
+    // Note there can only be a getter or only a setter, both are not required.
+    // Use the getter if available, or the setter if there is no getter.  
+    let loc: SourceLocation;
+    if (propertyObj.propertyType === 'accessor') {
+        if (propertyObj.getter) {
+            loc = propertyObj.getter.location
+        } else if (propertyObj.setter) {
+            loc = propertyObj.setter.location
+        }
+     } else {
+        loc = propertyObj?.dataProperty.location
+    }
     return stripKeysWithUndefinedVals({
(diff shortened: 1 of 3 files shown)


packages/lwc-language-server/src/javascript/type-mapping.ts
```diff
@@ -217,17 +217,25 @@ function getMemberProperty(propertyObj: ClassProperty): InternalClassMember | nu
...
 
+    // Note there can only be a getter or only a setter, both are not required.
+    // Use the getter if available, or the setter if there is no getter.  
+    let loc: SourceLocation;
+    if (propertyObj.propertyType === 'accessor') {
+        if (propertyObj.getter) {
+            loc = propertyObj.getter.location
+        } else if (propertyObj.setter) {
+            loc = propertyObj.setter.location
+        }
```
gbockus-sf, line 228: From local testing it appears there will be no property object without either a getter, setter, both, or a dataProperty.  The`externalToInternalLoc` method handle the case where this value is empty.

packages/lwc-language-server/src/javascript/type-mapping.ts
```diff
@@ -217,17 +217,25 @@ function getMemberProperty(propertyObj: ClassProperty): InternalClassMember | nu
...
         type: 'property',
         value,
         decorator: decoratorType,
         doc: propertyObj.__internal__doc,
-        loc: externalToInternalLoc(
-            propertyObj.propertyType === 'accessor'
-                ? propertyObj.getter.location
-                : propertyObj.dataProperty.location
-        ),
+        loc: externalToInternalLoc(loc),
```
gbockus-sf, line 238: Locally this project fails to compile for me.  Not sure what's up with that. 

image

CristiCanizales (approved):
LGTM. Thanks for the fix!
:shipit:

PR #587, opened 2024-02-12
feat: automatically updating `tsconfig.json`'s `paths` mapping for core workspaces.

rui-rayqiu:
### What does this PR do?

Impact: this PR only has impact on core workspaces for Salesforce internal users working on core. For external users working on SFDX workspaces, this PR should have no impact.

This PR adds a new feature for automatically updating `tsconfig.json`'s `paths` mapping for core workspaces.

LWC team is working on supporting TypeScript for authoring LWCs on core. In order for TypeScript compiler to work with LWC’s folder structure, we need to add a series of `paths` entries in `tsconfig.json` that re-map imports of LWC components to relative file paths of the imported components. For example, if component `one/cmpA` imports `force/wireUtils`, the import statement in `one/cmpA` will be like `import { someUtil } from ‘force/wireUtils’;` and for TypeScript to know where to look for `force/wireUtils` we need to add a path entry in `tsconfig.json` as the following.

```
    "compilerOptions": {
        "paths": {
            "force/wireUtils": ["../ui-force-components/modules/force/wireUtils/wireUtils"],
        }
    }
```

Maintaining this mapping in `tsconfig.json` manually is not convenient and a less ideal DX for developers on core. This PR adds a new indexer in the `lwc-language-server` that maintains a mapping between LWCs and their paths in `tsconfig.json`, automatically updating the `tsconfig.json` file when language server is initialized and whenever a LWC TypeScript file is changed. Thus developers won't need to create entries for `paths` manually when creating new LWCs in TypeScript.

This new feature will update `tsconfig.json` file for the `paths` attribute for all the core modules in the workspace when language server is initialized. Then whenever a LWC TS file is changed, all the imports are analyzed and any necessary LWC imports will be added to `tsconfig.json` automatically.

Here is a short demo [video link](https://drive.google.com/file/d/16JltE0-DKXjblkZqLajuRLRAUOlVxOLX/view) on this feature. In the video, the test workspace `lightning-language-server/test-workspace/core-like-workspace/coreTS/core` (created in this PR) is opened in VSCode, there are two modules `ui-force-components` and `ui-global-components` in the workspace. Both `tsconfig.json` files for them don't have a `paths` entries. After the indexer's init function is executed, both `tsconfig.json` files are updated with the `paths` mappings. After that, if I add an import statement in a ts file, the `tsconfig.json` file is also updated to have the mapping for the new import.

Testing this locally will require this [PR](https://github.com/forcedotcom/salesforcedx-vscode/pull/5423) for adding TS files to file watcher.

@W-14958700@

### DIFF
--- a/packages/lightning-lsp-common/src/__tests__/context.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/context.test.ts
@@ -154,6 +154,26 @@ it('isLWCJavascript()', async () => {
     expect(await context.isLWCJavascript(document)).toBeTruthy();
 });
 
+it('isLWCTypeScript()', async () => {
+    const context = new WorkspaceContext(CORE_PROJECT_ROOT);
+
+    // lwc .ts
+    let document = readAsTextDocument(CORE_PROJECT_ROOT + '/modules/one/app-nav-bar/app-nav-bar.ts');
+    expect(await context.isLWCTypeScript(document)).toBeTruthy();
+
+    // lwc .ts outside namespace root
+    document = readAsTextDocument(CORE_ALL_ROOT + '/ui-force-components/modules/force/input-phone/input-phone.ts');
+    expect(await context.isLWCTypeScript(document)).toBeFalsy();
+
+    // lwc .html
+    document = readAsTextDocument(CORE_PROJECT_ROOT + '/modules/one/app-nav-bar/app-nav-bar.html');
+    expect(await context.isLWCTypeScript(document)).toBeFalsy();
+
+    // lwc .js
+    document = readAsTextDocument(CORE_PROJECT_ROOT + '/modules/one/app-nav-bar/app-nav-bar.js');
+    expect(await context.isLWCTypeScript(document)).toBeFalsy();
+});
+
 it('configureSfdxProject()', async () => {
     const context = new WorkspaceContext('test-workspaces/sfdx-workspace');
     const jsconfigPathForceApp = FORCE_APP_ROOT + '/lwc/jsconfig.json';
--- a/packages/lightning-lsp-common/src/__tests__/context.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/context.test.ts
@@ -318,7 +318,7 @@ it('configureCoreMulti()', async () => {
     // verify newly created jsconfig.json
     verifyJsconfigCore(jsconfigPathGlobal);
     // verify jsconfig.json is not created when there is a tsconfig.json
-    expect(fs.existsSync(tsconfigPathForce)).not.toExist();
+    expect(fs.existsSync(jsconfigPathForce)).not.toExist();
     verifyTypingsCore();
 
     fs.removeSync(tsconfigPathForce);
--- a/packages/lightning-lsp-common/src/__tests__/context.test.ts
+++ b/packages/lightning-lsp-common/src/__tests__/context.test.ts
@@ -155,13 +155,14 @@ it('isLWCJavascript()', async () => {
 });
 
 it('isLWCTypeScript()', async () => {
+    // workspace root project is ui-global-components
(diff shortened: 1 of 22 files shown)


packages/lightning-lsp-common/src/__tests__/context.test.ts
```diff
@@ -154,6 +154,26 @@ it('isLWCJavascript()', async () => {
...
 
+it('isLWCTypeScript()', async () => {
+    const context = new WorkspaceContext(CORE_PROJECT_ROOT);
+
+    // lwc .ts
+    let document = readAsTextDocument(CORE_PROJECT_ROOT + '/modules/one/app-nav-bar/app-nav-bar.ts');
+    expect(await context.isLWCTypeScript(document)).toBeTruthy();
+
+    // lwc .ts outside namespace root
+    document = readAsTextDocument(CORE_ALL_ROOT + '/ui-force-components/modules/force/input-phone/input-phone.ts');
```
ravijayaramappa, line 165: Not sure I understand why this is outside of the namespace root.

packages/lwc-language-server/src/__tests__/typescript.test.ts
```diff
@@ -0,0 +1,205 @@
...
+                import {obj1} from './abc';
+                import {obj2} from '../xyz';
+                import {obj3} from 'lightning/confirm';
+                import {obj4} from '@salesforce/label/x';
+                import {obj5} from 'x.html';
+                import {obj6} from 'y.css';
+                import {obj7} from 'namespace/cmpName';
+            `);
+            const imports = await collectImportsForDocument(document);
+            expect(imports.size).toEqual(1);
```
ravijayaramappa, line 180: Is "lightning/confirm" excluded because it is an off-core component?

packages/lwc-language-server/src/typescript/tsconfig-path-indexer.ts
```diff
@@ -0,0 +1,321 @@
...
+                this.coreModulesWithTSConfig = this.getCoreModulesWithTSConfig(workspaceRoots);
+                break;
+        }
+    }
+
+    /**
+     * Gets all paths for TypeScript LWC components on core.
+     */
+    get componentEntries(): string[] {
+        const defaultSource = normalize(`${this.coreRoot}/*/modules/*/*/*.ts`);
```
ravijayaramappa, line 77: Assuming a folder layout like this
```
/ui-force-components
-->/modules
---->/force
------>/one
-------->/one.ts
-------->/util/util.ts
```

`/*/modules/*/*/*.ts` would not include any nested folders(`util.ts`) underneath a component folder. I think that is fine because the importer only can import the main entry file of a module.

packages/lightning-lsp-common/src/__tests__/context.test.ts
rui-rayqiu: This is to test that if a ts file that is outside of the VSCode workspace is modified, we do not update tsconfig.json. Here the context workspace root is `CORE_PROJECT_ROOT` which is `ui-global-components` so if a file in `ui-force-components` is modified the return value of `isLWCTypeScript` should return false. I'll add an inline comment to explain this.

ravijayaramappa (commented):
LGTM, I would defer to the language-server team to approve since there are aspects here that I am not familiar with.

packages/lwc-language-server/src/typescript/tsconfig-path-indexer.ts
rui-rayqiu: Yes, that's correct. We only want to include the main component entry ts file and ignore all other utility files.

nwcm:
Related but slightly off topic. This issue is also prevalent in SFDX workspaces and makes the DX painful. Would love to see similar ported to SFDX workspaces especially with the move to TS

PR #591, opened 2024-06-26
feat: configure `sfdx` workspace to write `tsconfig` files

jmsjtu:
### What does this PR do?

This PR adds the ability for the language server to generate a `tsconfig.json` in addition to the `jsconfig.json`.

The language server will generate a `tsconfig.sfdx.json` file in the `.sfdx` directory and a `tsconfig.json` in each `lwc` directory.

The feature is gated behind a feature flag that will be implemented in the `salesforcedx-vscode` repository.

@W-15644075@

### DIFF
--- a/packages/lightning-lsp-common/src/context.ts
+++ b/packages/lightning-lsp-common/src/context.ts
@@ -308,6 +308,14 @@ export class WorkspaceContext {
     }
 
     /**
+     * Configures LWC project to support TypeScript
+     */
+    public async configureProjectForTs(): Promise<void> {
+        // TODO: This should be moved into configureProject after dev preview
+        await this.writeTsconfigJson();
+    }
+
+    /**
      * Acquires list of absolute modules directories, optimizing for workspace type
      * @returns Promise
      */
@@ -444,6 +452,29 @@ export class WorkspaceContext {
                         const jsConfigPath = path.join(modulesDir, 'jsconfig.json');
                         this.updateConfigFile(jsConfigPath, jsConfigContent); // no workspace reference yet, that comes in update config file
                     }
+                }
+                break;
+        }
+    }
+
+    private async writeTsconfigJson(): Promise<void> {
+        switch (this.type) {
+            case WorkspaceType.SFDX:
+                // Write tsconfig.sfdx.json first
+                const baseTsConfigPath = path.join(this.workspaceRoots[0], '.sfdx', 'tsconfig.sfdx.json');
+                const baseTsConfig = await fs.readFile(utils.getSfdxResource('tsconfig-sfdx.base.json'), 'utf8');
+                this.updateConfigFile(baseTsConfigPath, baseTsConfig);
+                // Write to the tsconfig.json in each module subdirectory
+                const tsConfigTemplate = await fs.readFile(utils.getSfdxResource('tsconfig-sfdx.json'), 'utf8');
+                const forceignore = path.join(this.workspaceRoots[0], '.forceignore');
+                // TODO: We should only be looking through modules that have TS files
+                const modulesDirs = await this.getModulesDirs();
+                for (const modulesDir of modulesDirs) {
+                    const tsConfigPath = path.join(modulesDir, 'tsconfig.json');
+                    const relativeWorkspaceRoot = utils.relativePath(path.dirname(tsConfigPath), this.workspaceRoots[0]);
+                    const tsConfigContent = this.processTemplate(tsConfigTemplate, { project_root: relativeWorkspaceRoot });
+                    this.updateConfigFile(tsConfigPath, tsConfigContent);
+                    await this.updateForceIgnoreFile(forceignore);
                 }
                 break;
         }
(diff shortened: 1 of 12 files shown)


packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -130,6 +131,15 @@ export default class Server {
         };
     }
 
+    async onInitialized(): Promise<void> {
+        // The config value comes from salesforcedx-vscode/salesforcedx-vscode-lwc
+        const hasTsEnabled = await this.connection.workspace.getConfiguration('salesforcedx-vscode-lwc.preview.typeScriptSupport');
```
jmsjtu, line 136: The entire feature is gated behind this feature flag, which hasn't been implemented in salesforcedx-vscode yet.

There is a follow-up PR to implement the flag, in the meantime, I've tested locally that when the config value is enabled it works correctly.

packages/lwc-language-server/src/lwc-server.ts
```diff
@@ -80,6 +80,7 @@ export default class Server {
 
     constructor() {
         this.connection.onInitialize(this.onInitialize.bind(this));
+        this.connection.onInitialized(this.onInitialized.bind(this));
```
jmsjtu, line 83: Looks like config values are not available during `onInitialize` which is why I'm using `onInitialized` instead.

Once the feature is GA, we can move everything back to `onInitialize`.

packages/lightning-lsp-common/src/context.ts
```diff
@@ -449,6 +457,29 @@ export class WorkspaceContext {
...
+            case WorkspaceType.SFDX:
+                // Write tsconfig.sfdx.json first
+                const baseTsConfigPath = path.join(this.workspaceRoots[0], '.sfdx', 'tsconfig.sfdx.json');
+                const baseTsConfig = await fs.readFile(utils.getSfdxResource('tsconfig-sfdx.base.json'), 'utf8');
+                this.updateConfigFile(baseTsConfigPath, baseTsConfig);
+                // Write to the tsconfig.json in each module subdirectory
+                const tsConfigTemplate = await fs.readFile(utils.getSfdxResource('tsconfig-sfdx.json'), 'utf8');
+                const forceignore = path.join(this.workspaceRoots[0], '.forceignore');
+                // TODO: We should only be looking through modules that have TS files
+                const modulesDirs = await this.getModulesDirs();
```
jmsjtu, line 471: In a follow-up PR I plan to only check for `lwc` directories that actually have TS files.

For now, I'm writing to all `lwc` directories.

packages/lightning-lsp-common/src/resources/sfdx/tsconfig-sfdx.base.json
```diff
@@ -0,0 +1,13 @@
+{
+    "typeAcquisition": {
+        "include": [
+            "jest"
+        ]
+    },
```
wjhsf, line 6:
```suggestion
```

> Type Acquisition is only important for JavaScript projects. In TypeScript projects you need to include the types in your projects explicitly. However, for JavaScript projects, the TypeScript tooling will download types for your modules in the background and outside of your node_modules folder.

As per the [TypeScript docs](https://www.typescriptlang.org/tsconfig/#typeAcquisition), `typeAcquisition` isn't needed for TypeScript projects.

packages/lwc-language-server/src/lwc-server.ts
jmsjtu: After discussing with the team, we should remove the `tsconfig` if `salesforcedx-vscode-lwc.preview.typeScriptSupport = false`.

packages/lightning-lsp-common/src/context.ts
jmsjtu: We should also only be generating either a `jsconfig.json` or a `tsconfig.json` but not both, this can be done once we've moved to GA.

mingxuanzhangsfdx:
NIT: In order to pass the step `validate pr` in CI you have to follow the format in the description

> What issues does this PR fix or reference?
@[W-The work item ID]@

packages/lightning-lsp-common/src/context.ts
```diff
@@ -449,6 +457,29 @@ export class WorkspaceContext {
...
+                for (const modulesDir of modulesDirs) {
+                    const tsConfigPath = path.join(modulesDir, 'tsconfig.json');
+                    const relativeWorkspaceRoot = utils.relativePath(path.dirname(tsConfigPath), this.workspaceRoots[0]);
+                    const tsConfigContent = this.processTemplate(tsConfigTemplate, { project_root: relativeWorkspaceRoot });
+                    this.updateConfigFile(tsConfigPath, tsConfigContent);
+                    await this.updateForceIgnoreFile(forceignore);
```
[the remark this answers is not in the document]
mingxuanzhangsfdx: nvm I saw the PR you created

mingxuanzhangsfdx (approved):
Looks good!

packages/lightning-lsp-common/src/context.ts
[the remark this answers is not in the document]
jmsjtu: This is the same behavior as with `jsconfig.json`, if the language server detects a `tsconfig.json` it will add it to the ignore file.

PR #594, opened 2024-07-09
feat: automatically update `tsconfig.sfdx.json`'s paths 

jmsjtu:
### What does this PR do?

This PR adds the ability for the language server to update the `paths` mapping in `tsconfig.sfdx.json`.

The purpose for doing this is to allow TypeScript to determine where to locate an LWC component.

For example:

```js
import Foo from 'c/bar';
...
```

TypeScript doesn't know how to resolve`c/bar` and requires it to be mapped in the `tsconfig.json`, ex:

```js
{
    "compilerOptions": {
        "paths": {
            "c/*": ["/user/some/where/*/bar"]
        }
    }
}
```

The path mapping is updating in 2 scenarios:

1. When the language server first starts.
2. When there is a file change to a js or ts LWC component.

What issues does this PR fix or reference?
@W-15572824@

### DIFF
--- a/packages/lightning-lsp-common/src/utils.ts
+++ b/packages/lightning-lsp-common/src/utils.ts
@@ -4,7 +4,7 @@ import { TextDocument, FileEvent, FileChangeType } from 'vscode-languageserver';
 import { URI } from 'vscode-uri';
 import equal from 'deep-equal';
 import { WorkspaceContext } from './context';
-import { WorkspaceType, isLWC } from './shared';
+import { WorkspaceType } from './shared';
 import { promisify } from 'util';
 import { Glob } from 'glob';
 import * as jsonc from 'jsonc-parser';
--- a/packages/lightning-lsp-common/src/utils.ts
+++ b/packages/lightning-lsp-common/src/utils.ts
@@ -1,10 +1,10 @@
 import * as fs from 'fs-extra';
-import { basename, extname, join, relative, resolve } from 'path';
+import { basename, extname, join, parse, relative, resolve } from 'path';
 import { TextDocument, FileEvent, FileChangeType } from 'vscode-languageserver';
 import { URI } from 'vscode-uri';
 import equal from 'deep-equal';
 import { WorkspaceContext } from './context';
-import { WorkspaceType } from './shared';
+import { WorkspaceType, isLWC } from './shared';
 import { promisify } from 'util';
 import { Glob } from 'glob';
 import * as jsonc from 'jsonc-parser';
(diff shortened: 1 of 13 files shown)


packages/lightning-lsp-common/src/utils.ts
```diff
@@ -74,6 +74,23 @@ export async function includesDeletedAuraWatchedDirectory(context: WorkspaceCont
...
+        if (event.type !== FileChangeType.Changed) {
+            const insideLwcWatchedDirectory = await isLWCWatchedDirectory(context, event.uri);
+            if (insideLwcWatchedDirectory) {
+                const file = toResolvedPath(event.uri);
+                const { dir, name, ext } = parse(file);
+                // Only update path mapping for newly created lwc modules
+                if (/.*(.ts|.js)$/.test(ext) && dir.endsWith(name)) {
+                    return true;
+                }
+            }
```
jmsjtu, line 88: The criteria for updating the path mapping is as follows:

1. The file change event was a creation or deletion.
2. The file change is a `.js` or `.ts` file.
3. The directory of the change matches the file's name, ex: foo/foo.js
4. The file's parent directory is an lwc directory, ex: lwc/foo/foo.js

packages/lwc-language-server/src/__tests__/component-indexer.test.ts
```diff
@@ -125,6 +126,54 @@ describe('ComponentIndexer', () => {
...
+                'force-app/main/default/lwc/*/import_relative',
+                'force-app/main/default/lwc/*/index',
+                'force-app/main/default/lwc/*/lightning_datatable_example',
+                'force-app/main/default/lwc/*/lightning_tree_example',
+                'force-app/main/default/lwc/*/todo_item',
+                'force-app/main/default/lwc/*/todo',
+                'force-app/main/default/lwc/*/typescript',
+                'force-app/main/default/lwc/*/utils',
+                'utils/meta/lwc/*/todo_util',
+                'utils/meta/lwc/*/todo_utils',
```
jmsjtu, line 142: This is a sample of how the `paths` mapping will look like, it includes both `.js` and `.ts` files so TypeScript can resolve all LWC module locations.

packages/lwc-language-server/src/component-indexer.ts
```diff
@@ -122,6 +122,43 @@ export default class ComponentIndexer extends BaseIndexer {
...
+                console.error(err);
+            }
+        }
+    }
+
+    get tsConfigPathMappingFiles(): string[] {
+        let files: string[] = [];
+        if (this.workspaceType === WorkspaceType.SFDX) {
+            const sfdxSource = normalize(`${this.workspaceRoot}/${this.sfdxPackageDirsPattern}/**/*/lwc/*/*.{js,ts}`);
+            const filePaths = sync(sfdxSource, { stats: true });
```
jmsjtu, line 148: This does a fast glob search through the file system with the given glob pattern.

Note that we're not reading the actual files, just scanning the file system.

This will be done each time the mapping needs to be updated.

Performance should be fine in this scenario, I believe even with a few thousand components the search should only take a new miliseconds.

packages/lightning-lsp-common/src/utils.ts
jmsjtu: This logic is actually missing a scenario where a directory is deleted, the change event only includes the path to the directory which means it will fail the check.

I will push a commit to fix it soon.

packages/lwc-language-server/src/__tests__/component-indexer.test.ts
```diff
@@ -125,6 +126,54 @@ describe('ComponentIndexer', () => {
...
+
+            describe('#tsConfigPathMappingFiles', () => {
+                it('returns a list of files where the .js or.ts filename is the same as its parent directory name', () => {
+                    const tsConfigPathMapping = componentIndexer.tsConfigPathMappingFiles.sort();
+                    expect(tsConfigPathMapping).toEqual(expectedComponents.sort());
+                });
+            });
+
+            describe('updateSfdxTsConfigPath', () => {
+                it('updates updates tsconfig.sfdx.json path mapping', async () => {
```
daphne-sfdc, line 157: This looks like a typo.

```suggestion
                it('updates tsconfig.sfdx.json path mapping', async () => {
```

daphne-sfdc (approved):
Looks good to me! 🎉 

All unit tests pass and manual testing works as expected ✅ 

Just one typo (see above comment) then this is good to be merged

Test plan here: https://salesforce.quip.com/5idtApGvFMUT

PR #601, opened 2024-09-03
fix(forceignore): only add tsconfig + ts to forceignore for TS projects @W-16625817@ 

wjhsf:
### What does this PR do?

This PR updates the language server to only add TypeScript-related files to the `.forceignore` if the project has TypeScript support enabled. It also sets an additional option in the base tsconfig that is required for projects to correctly resolve modules.

Fixes https://github.com/forcedotcom/salesforcedx-vscode/issues/5802

@W-16625817@

### DIFF
--- a/packages/lightning-lsp-common/src/resources/sfdx/tsconfig-sfdx.base.json
+++ b/packages/lightning-lsp-common/src/resources/sfdx/tsconfig-sfdx.base.json
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "skipLibCheck": true,
         "target": "ESNext",
+        "module": "NodeNext",
         "paths": {
             "c/*": []
         }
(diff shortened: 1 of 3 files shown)


packages/lightning-lsp-common/src/resources/sfdx/tsconfig-sfdx.base.json
```diff
@@ -2,6 +2,7 @@
     "compilerOptions": {
         "skipLibCheck": true,
         "target": "ESNext",
+        "module": "NodeNext",
```
mingxuanzhangsfdx, line 5: It looks like updating the node version causes existing unit tests failures, which is a WI we are going to support. The latest node version unit tests support is node v18. Can you revert this for the tests to pass by any chance?

wjhsf: This is not updating the node version, this is specifying the TypeScript module resolution algorithm. The tests were failing because I missed a line when asserting the contents of the JSON file.

mingxuanzhangsfdx (approved):
Thanks for the change. Test case:

1. dev preview feature disabled: forceignore is not updated.
2. enable dev preview feature: able to see the update of ts and tsconfig in forceignore.
3. There is an extra entry "module: 'NodeNext'" in .sfdx/tsconfig.sfdx.json

PR #615, opened 2025-08-21
fix: removing lodash as it won't be needed in DX

madhur310:
### What does this PR do?
- remove lodash/lodash.template dependencies. Uses cjs instead which is already available in DX repo.
- fixes unrelated lint error.
- 

@W-19377429@


packages/lightning-lsp-common/src/context.ts
```diff
@@ -39,23 +37,6 @@ async function findSubdirectories(dir: string): Promise<string[]> {
...
-/**
- * @return list of .js modules inside namespaceRoot folder
- */
-async function findModulesIn(namespaceRoot: string): Promise<string[]> {
-    const files: string[] = [];
-    const subdirs = await findSubdirectories(namespaceRoot);
-    for (const subdir of subdirs) {
-        const basename = path.basename(subdir);
-        const modulePath = path.join(subdir, basename + '.js');
-        if ((await fs.pathExists(modulePath)) && componentUtil.isJSComponent(modulePath)) {
```
mshanemc, line 51: removing this made componentUtil unused import.

can you adjust the lint/ts rules to prevent that kind of thing?

mshanemc (changes requested):
remove unused code

PR #624, opened 2025-09-30
fix: prevent publish test-workspaces W-19605327

madhur310:
[@W-19605327@](https://gus.lightning.force.com/a07EE00002LWVSwYAP)


packages/aura-language-server/.npmignore
mshanemc, line 1: Huh.  We don't normally use `npmignore`.  The normal approach is to have a `files` property on the package.json that  says what files go in, instead of trying to explicitly ignore files.

mshanemc: ex: https://github.com/forcedotcom/sfdx-core/blob/69c0055e7303decdc6d4d79d4ea1944f7e90287a/package.json#L50

I'd rather have "e2e failed because you forgot to put something in `package.json#files`" rather than, "you accidentally package unnecessary stuff because you didn't put it in the `.npmignore` file"

packages/lightning-lsp-common/src/resources/core/core.code-workspace.json
```diff
@@ -38,6 +38,6 @@
         "extensions.ignoreRecommendations": true,
         "salesforcedx-vscode-lightning.activationMode": "always",
         "salesforcedx-vscode-lightning.showLightningExplorer": true,
-        "salesforcedx-vscode-core.internal-development": true
```
mshanemc, line 41: are these sneaking in from writing base-context?

https://github.com/forcedotcom/lightning-language-server/blob/aaadd02c2c199929b721766545317d9c654814d6/packages/lightning-lsp-common/src/base-context.ts#L261-L262

mshanemc: maybe we're missing some kind of check to verify that you're working in the salesforce internal repo that all this supports?

mshanemc: the big problem with making changes is that we don't test that scenario at all (using ext for internal development) but there's probably users that need it.

madhur310:
closing after further discussion offline, we will be moving things to mono repo anyways.

